PR #16, opened 2016-07-08
Correct the format of the README in UI directory

SLMT:
I found that the README file in the UI directory is not able to be correctly rendered on Github. I guessed you might want to put a table in the file.

So I corrected the format of the table in order to make it display correctly.


shepmaster:
Thanks! I always screw up Markdown tables, and I forgot to check this file 😊

PR #48, opened 2016-08-07
Keep track of CWD in README

Byron:
This is analogue to what was already done in the `Build the containers` section.


shepmaster:
Thanks! That was a bit sloppy of me ☺️

PR #49, opened 2016-08-07
Make `--pids-limit` a default feature

Byron:
That way it is possible to not use this flag by building like so:

`cargo build --no-default-features`

This is useful for local sandbox installations running on a different
OS which doesn't directly support limiting PIDs, or which works
differently in that regards.

I am using coreOS for example, but admittedly also didn't try
to find a way to make it work.

### DIFF
--- a/ui/Cargo.toml
+++ b/ui/Cargo.toml
@@ -4,6 +4,10 @@ name = "ui"
 version = "0.1.0"
 build = "build.rs"
 
+[features]
+default = ['fork-bomb-prevention']
+fork-bomb-prevention = []
+
 [build-dependencies]
 serde_codegen = "0.7.0"
 
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -199,10 +199,13 @@ impl Sandbox {
             .args(&["--net", "none"])
             .args(&["--memory", "256m"])
             .args(&["--memory-swap", "320m"])
-            .args(&["--pids-limit", "512"])
             .args(&["--env", "PLAYGROUND_TIMEOUT=10"])
             .args(&["--env", "RUST_NEW_ERROR_FORMAT=true"]);
 
+        if cfg!(feature = "fork-bomb-prevention") {
+            cmd.args(&["--pids-limit", "512"]);
+        }
+        
         cmd
     }
 }


ui/src/sandbox.rs
```diff
@@ -190,19 +190,23 @@ impl Sandbox {
...
-        cmd
-            .arg("run")
+        cmd.arg("run")
             .arg("--rm")
-            .arg("--volume").arg(&mount_input_file)
-            .arg("--volume").arg(&mount_output_dir)
+            .arg("--volume")
+            .arg(&mount_input_file)
+            .arg("--volume")
+            .arg(&mount_output_dir)
```
shepmaster: Could you undo these changes? I'm attempting to keep it as `name, value` to more easily match up what it does.

ui/Cargo.toml
```diff
@@ -4,6 +4,10 @@ name = "ui"
 version = "0.1.0"
 build = "build.rs"
 
+[features]
+default = ['pids-limit']
+pids-limit = []
```
shepmaster: Maybe add a comment here along the lines of:

```toml
# This feature prevents forkbombs
```

To make it blatantly obvious to anyone disabling it what issues they may be opting into.

ui/src/sandbox.rs
Byron: Will do ! Sorry for that, it was a rustfmt slip-up. Had it enabled on file-save, which mangled the entire file. When I noticed that, I copied the entire function to keep my changes after checking the file back out. Lesson learned: Watch these diffs :) !

ui/Cargo.toml
Byron: I did it by changing the name of the feature. What do you think ?

Byron:
Alright, thanks for the feedback ! Please have a look to see whether the comments of been addressed.

shepmaster:
Great! I appreciate that you renamed the feature instead of just adding a comment.

Would you mind squashing the 3 commits into one? If you'd rather, I can do it on my side.

Byron:
Squash successful ! Thanks for merging.

shepmaster:
For your own sanity, you may want to figure out how to enable PID limiting anyway.

On Alpine Linux (for EC2), I had to do something like:

```
vim /etc/cgconfig.conf
# Add:    pids       = /cgroup/pids;
```

To mount the appropriate cgroup. Having an up-to-date kernel and Docker was also useful. I'd be surprised if CoreOS doesn't support this in some manner.

Byron:
Indeed, for a production system I wouldn't want to have it any other way either.

By the way, for inspiration I recommend having a look at the *Crystal Playground* implementation. You can run it simply using `crystal play`, and it looks like this:

screen shot 2016-08-07 at 16 48 46

PR #53, opened 2016-08-23
Standardize font sizes in the body of the playground

shepmaster:


shepmaster:
Before

image

After

image

PR #114, opened 2017-03-16
CORS support

atorstling:
Hi!

Great work with the playground! I wanted to use the playground to power interactiveness in my presentation slides (like in the official Rust docs). That required connecting to the UI server from pages hosted on another domain. That didn't work out so well without CORS. So I added support. I made it really open (*) and therefore put it behind a feature toggle: envvar "CORS". What do you think? Thanks for considering.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,5 @@
+Thanks for taking the time to file an issue! 
+
+If you are requesting that a specific crate be added, please review [our policy][policy] about which crates are accepted. TL;DR: we don't currently accept one-off additions of crates.
+
+[policy]: https://github.com/integer32llc/rust-playground/issues/101


ui/src/main.rs
```diff
@@ -72,10 +79,45 @@ fn main() {
...
+        Box::new(move | req: &mut Request | {
+            match req.method {
+                Method::Options => Ok(Response::with((status::Ok))),
+                _ => handler.handle(req)
+            }
+        })
+    }
+}
+
+impl AfterMiddleware for CorsMiddleware {
```
shepmaster: Is there a reason to use an "after" middleware at all? What if in the "around" middleware, we simply constructed the response with the proper headers to start with?

shepmaster: Or do these headers need to be on *every* response?

shepmaster: Maybe just the `AccessControlAllowOrigin::Value("*".to_string())` needs to be on every request?

ui/src/main.rs
```diff
@@ -72,10 +79,45 @@ fn main() {
...
+        let headers = vec![
+                    UniCase("Origin".to_owned()),
+                    UniCase("X-Requested-With".to_owned()),
+                    UniCase("Content-Type".to_owned()),
+                    UniCase("Accept".to_owned()),
+                    ];
+        resp.headers.set(AccessControlAllowHeaders(headers));
+        let methods = vec![ Method::Get, Method::Put, Method::Post ];
+        resp.headers.set(AccessControlAllowMethods(methods));
+        Ok(resp)
```
shepmaster: We should consider adding `Access-Control-Max-Age` with 1 hour or so, to avoid the browser having to do the preflight checks over and over.

shepmaster:
Additionally, do you have the source code for the presentation that uses this available somewhere? I'd love to browse it (and maybe use it for my own purposes!)

shepmaster:
There's also https://crates.io/crates/iron-cors, which I haven't investigated

atorstling:
Hello! Truth be told I didn't really put much effort into making a minimal and "correct" implementation of the feature. I see you have high standards, so I'll do some investigations and get back to you. Also see my inline comments. My presentation is really just a brief presentation of Rust basics for newcomers, but feel free to browse it at https://github.com/atorstling/rusttalk . Cheers!

ui/src/main.rs
```diff
@@ -72,10 +79,45 @@ fn main() {
...
     Iron::new(chain).http((&*address, port)).expect("Unable to start server");
 }
 
+struct CorsMiddleware;
+
+impl AroundMiddleware for CorsMiddleware {
+    fn around(self, handler: Box<Handler>) -> Box<Handler> {
+        Box::new(move | req: &mut Request | {
+            match req.method {
+                Method::Options => Ok(Response::with((status::Ok))),
```
atorstling: Yes, you're right. Ideally one would also check if the route is already handling I can add a guard here to only respond to request with such headers. It might be worthwhile to not respond ahead of time for routes who are already OPTION requests, but I am unsure if it's possible to identify those at this stage. I also noticed that you are using Mount and not Router, so I assume that all routes accept all HTTP methods? I

ui/src/main.rs
atorstling: I figured that the easiest way of adding the headers was an "after" middleware. Later I added the "around" middleware for intercepting options requests. But if you prefer having only the more powerful "around" middleware, I can change this. It makes more sense if we'll detect the preflight request. And I think that you are correct that only alloworigin is required on non-preflight requests. I'll do some testing and get back to you.

ui/src/main.rs
```diff
@@ -72,10 +79,45 @@ fn main() {
...
+}
+
+impl AfterMiddleware for CorsMiddleware {
+    fn after(&self, _: &mut Request, mut resp: Response) -> IronResult<Response> {
+        resp.headers.set(AccessControlAllowOrigin::Value("*".to_string()));
+        let headers = vec![
+                    UniCase("Origin".to_owned()),
+                    UniCase("X-Requested-With".to_owned()),
+                    UniCase("Content-Type".to_owned()),
+                    UniCase("Accept".to_owned()),
```
atorstling: No, I will investigate a more conservative setting. This was one of "the most generous" option I found quoted on the web. I'm not sure what you would prefer, should we be generous or restrictive?

ui/src/main.rs
```diff
@@ -72,10 +79,45 @@ fn main() {
...
+                    ];
+        resp.headers.set(AccessControlAllowHeaders(headers));
+        let methods = vec![ Method::Get, Method::Put, Method::Post ];
```
atorstling: Just a copy-paste. I'll remove the put.

ui/src/main.rs
atorstling: Good idea! Didn't know about that one.

ui/src/main.rs
atorstling: Accept seems unnecessary. 
The Content-Type header seems to be required for application/json. ( from https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Headers). X-requested-with seems to be added by default by some js frameworks: http://stackoverflow.com/questions/17478731/whats-the-point-of-the-x-requested-with-header . Origin might be superflous.

I took these from https://enable-cors.org/server_expressjs.html , couldn't find any rationale for their chose there.

ui/src/main.rs
atorstling: I think I'll use this flowchart for detecting, seems legit :) : https://www.html5rocks.com/static/images/cors_server_flowchart.png

PR #120, opened 2017-03-23
Slim down the bundle size

shepmaster:


shepmaster:
Travis doesn't have the newest nightly. Let's roll the dice...

Issue #101, opened 2017-02-11
Consider using downloads in the last N days, instead of all time.


shepmaster:
See discussion started in #82

sgrif:
All I'm saying is... If you made libsqlite3 available in the sandbox, we would find it super helpful for bug reports. :)

ArtemGr:
Will the "last N days" selection help libsqlite3? How would one check that?

gsingh93:
Only somewhat related to this issue and the discussion in #82, but it seems kind of strange (if I'm understanding this correctly) that updating the top crates list could cause code that previously used to work to no longer work. i.e. the 100th crate in the list may drop to 101, the service is redeployed, and now any code snippets that relied on that crate no longer work. That being said, I don't know what the best solution would be.

shepmaster:
> 100th crate in the list may drop to 101

Even easier, a crate can simply release a new backwards-incompatible version. Since we always use the more-or-less-latest version, this is quite possible.

> any code snippets that relied on that crate no longer work

This is a good point, but I don't see any good solution unless some benefactor decides to provide financial contribution ;-). 

The *Big Solution* I can think of is to allow arbitrary versions of crates (and we might was well throw in arbitrary Rust versions at that point). Then we are one step away from being a (poor facsimile) of a web-based IDE.

> no longer work

Beneficially, any *code* is still available, even if it's not executable.

shepmaster:
> code is still available

Which is part of the reason that I avoid adding the shortlink button and prefer the gist button — I'm betting that GitHub and Gists will last a long time!

shepmaster:
Looks like the last 90 days right now would be 

```
  id  |          name           | downloads 
------+-------------------------+-----------
  463 | serde                   |   1387484
  795 | libc                    |    880648
  793 | bitflags                |    713088
  363 | lazy_static             |    541543
  524 | rustc-serialize         |    536186
  429 | winapi                  |    519486
  547 | log                     |    500864
 2164 | winapi-build            |    487439
 1339 | rand                    |    487066
  811 | kernel32-sys            |    486698
   35 | gcc                     |    451679
 4746 | num-traits              |    430125
 4697 | thread_local            |    427619
 2233 | regex-syntax            |    423146
 4446 | thread-id               |    413968
 2365 | aho-corasick            |    412091
 2364 | memchr                  |    403324
 1592 | num_cpus                |    400180
  545 | regex                   |    399375
 3231 | utf8-ranges             |    391706
[...]

PR #133, opened 2017-04-13
Add futures crate to playground.

withoutboats:
I don't know what your policy is for adding more crates but it seems like a good idea to have futures since it underpins the async ecosystem.


shepmaster:
> I don't know what your policy is

[From the README](https://github.com/integer32llc/rust-playground#whats-different):

> The top 100 popular crates (ranked by all-time downloads) and their dependencies are available for use

Closing as a dupe of #101.

PR #136, opened 2017-04-25
Update top crates to pick up Serde 1.0

dtolnay:
@shepmaster `cargo tree -d` shows that there are no top crates still using serde 0.9.


shepmaster:
Thanks!

PR #138, opened 2017-05-10
launch sandbox with --cap-drop=ALL and --security-opt=no-new-privileges

dwrensha:
Adding these arguments should make it significantly more difficult to break out of the sandbox.

Documentation links:
 - https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities
 - http://www.projectatomic.io/blog/2016/03/no-new-privs-docker/

Note that play.rust-lang.org [already uses these options](https://github.com/rust-lang/rust-playpen/blob/ac799f6f69389ca68bbc52f46a8956938c06317e/src/docker.rs#L22-L26).

### DIFF
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -217,6 +217,8 @@ impl Sandbox {
             .arg("--rm")
             .arg("--volume").arg(&mount_input_file)
             .arg("--volume").arg(&mount_output_dir)
+            .arg("--cap-drop=ALL")
+            .arg("--security-opt=no-new-privileges")
             .args(&["--workdir", "/playground"])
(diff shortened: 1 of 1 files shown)


shepmaster:
Thanks!

shepmaster:
Hmm, this actually breaks some features. For example, formatting fails with:

> Error writing files: Permission denied (os error 13)

Adding `--cap-add=DAC_OVERRIDE` allows them to work again. Any objections to adding that capability back?

dwrensha:
Interesting! I wonder if this has to do with some interaction with selinux.

Sure, add it back; dropping all capabilities except that one is still better than keeping all capabilities.

PR #152, opened 2017-05-22
Cors Again

atorstling:
This is a sequel to #114 . After your suggestions I had a look at existing CORS libraries for Iron to see if I could use them to make a more correct implementation of CORS for the playground. I concluded that there was definitely room for a more ambitious CORS library, so I totally rabbit-holed and went ahead and wrote it myself. 

In order to minimize the amount of CORS code in the playground, I used a single config for all endpoints.

### DIFF
--- a/ui/src/main.rs
+++ b/ui/src/main.rs
@@ -14,6 +14,7 @@ extern crate serde_derive;
 extern crate mktemp;
 #[macro_use]
 extern crate quick_error;
+extern crate corsware;
 
 use std::any::Any;
 use std::convert::{TryFrom, TryInto};
@@ -25,6 +26,8 @@ use iron::headers::ContentType;
 use iron::modifiers::Header;
 use iron::prelude::*;
 use iron::status;
+use iron::method::Method::{Get, Post};
+use corsware::{CorsMiddleware, AllowedOrigins, UniCase};
 
 use mount::Mount;
 use serde::Serialize;
@@ -41,6 +44,7 @@ const DEFAULT_LOG_FILE: &'static str = "access-log.csv";
 
 mod sandbox;
 
+const ONE_HOUR_IN_SECONDS: u32 = 60 * 60;
 const ONE_DAY_IN_SECONDS: u64 = 60 * 60 * 24;
 const ONE_YEAR_IN_SECONDS: u64 = 60 * 60 * 24 * 365;
 
@@ -51,6 +55,7 @@ fn main() {
     let address = env::var("PLAYGROUND_UI_ADDRESS").unwrap_or(DEFAULT_ADDRESS.to_string());
     let port = env::var("PLAYGROUND_UI_PORT").ok().and_then(|p| p.parse().ok()).unwrap_or(DEFAULT_PORT);
     let logfile = env::var("PLAYGROUND_LOG_FILE").unwrap_or(DEFAULT_LOG_FILE.to_string());
+    let cors_enabled = env::var_os("PLAYGROUND_CORS_ENABLED").is_some();
 
     let files = Staticfile::new(&root).expect("Unable to open root directory");
     let mut files = Chain::new(files);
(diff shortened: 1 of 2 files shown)


ui/src/main.rs
```diff
@@ -25,6 +26,8 @@ use iron::headers::ContentType;
 use iron::modifiers::Header;
 use iron::prelude::*;
 use iron::status;
+use iron::method::Method::{Get,Post};
```
shepmaster: Should have space after the comma.

ui/src/main.rs
```diff
@@ -72,9 +75,19 @@ fn main() {
     let file_logger = FileLogger::new(logfile).expect("Unable to create file logger");
     let logger = StatisticLogger::new(file_logger);
     let rewrite = Rewrite::new(vec![vec!["help".into()]], "/index.html".into());
+    let cors = CorsMiddleware { 
+        allowed_origins : AllowedOrigins::Any { allow_null: false },
```
shepmaster: Shouldn't have a space before the `:`

ui/src/main.rs
```diff
@@ -72,9 +75,19 @@ fn main() {
...
     let logger = StatisticLogger::new(file_logger);
     let rewrite = Rewrite::new(vec![vec!["help".into()]], "/index.html".into());
+    let cors = CorsMiddleware { 
+        allowed_origins : AllowedOrigins::Any { allow_null: false }, 
+        allowed_headers: vec![UniCase("Content-Type".to_owned())],
+        allowed_methods : vec![ Get, Post ],
+        exposed_headers: vec![],
+        allow_credentials: false,
+        max_age_seconds: 60 * 60,
+        prefer_wildcard: true
```
shepmaster: Should have a trailing comma.

ui/src/main.rs
```diff
@@ -72,9 +75,19 @@ fn main() {
     let file_logger = FileLogger::new(logfile).expect("Unable to create file logger");
     let logger = StatisticLogger::new(file_logger);
     let rewrite = Rewrite::new(vec![vec!["help".into()]], "/index.html".into());
+    let cors = CorsMiddleware { 
+        allowed_origins : AllowedOrigins::Any { allow_null: false }, 
+        allowed_headers: vec![UniCase("Content-Type".to_owned())],
+        allowed_methods : vec![ Get, Post ],
```
shepmaster: Ditto space before `:`; no space inside the `[]`.

ui/src/main.rs
```diff
@@ -72,9 +75,19 @@ fn main() {
...
+        exposed_headers: vec![],
+        allow_credentials: false,
+        max_age_seconds: 60 * 60,
```
shepmaster: Just as a note for your crate, you might want to take a `std::time::Duration` here; then you wouldn't have to add `seconds` to the field name.

atorstling:
No worries about the old PR. I wasn't really sure how much time I would have and so I didn't want to advertise my plan up front if it wouldn't go through, so I just closed the PR. I should probably have said something though, sorry about that.

I'd like to fix everything in this PR. In fact, I was just about to squash before I pushed it, but I thought I would go through review  first :) 

I agree with most of your comments. The only thing I'm a bit unsure about is the Duration change. I appreciate you pointing out the existence of a std time lib, but I don't really like that the resolution of the duration would be lost. There is a risk that users would set a few hundred ms duration and not really understand why that didn't work as intended. That could of course be remedied with docs. But I'm not entirely sure that would be better than the current situation. What's your opinion on that?

atorstling:
Alright, how does this update look?

shepmaster:
> but I don't really like that the resolution of the duration would be lost. There is a risk that users would set a few hundred ms duration and not really understand why that didn't work as intended. That could of course be remedied with docs. But I'm not entirely sure that would be better than the current situation. What's your opinion on that?

Yeah, I think that your point is well-made. A note in the `Duration` docs says:

>  APIs binding a system timeout will typically round up the nanosecond precision **if the underlying system does not support that level of precision**.

I think you could make the same argument, just that your precision is one second (I assume).

However, all that being said, the change looks great to me, so merge ahoy!

atorstling:
Nice! I looked up the precision of the Max-Age param, and it's indeed one second: https://tools.ietf.org/html/rfc2616#section-3.3.2 . I agree with you that there is almost always a need for rounding when using a Duration, but I think the expectation is at least in the ms resolution range. You could make a case  for using it, however not doing so also simplified the implementation. I'll keep it in mind for the future.

Anyways, thanks for the merge!

shepmaster:
FYI, this repo got promoted to the official Rust Playground. It turns out that the Rust homepage and the books both use a *seeekret* route to run code on the page. Guess what that needs? That's right: CORS! Thank you so much for adding this support, which made responding to those requests **way** easier!

atorstling:
Hey, that's great news! From what I've seen, your playground seems like a definitive upgrade! 

I spotted that secret route with Chrome devtools, that's basically what started me down this route of adding CORS. Lucky coincidence that this should happen now. 

Anyway, great that I could help and also nice that other people can benefit from this change. Let me know if there is something you would like improved in the lib if you encounter something.

Issue #113, opened 2017-03-13
No `rayon` crate in the playground


matklad:
Hi! 

It seems like the `rayon` crate is missing: http://play.integer32.com/?gist=04382a80d3f1f36a60a9a4933d382d92&version=stable

I think it should be included, it's one of the nicest crates to demonstrate the power of Rust :)

shepmaster:
Sorry! We strive to have simple heuristics to decide which crates are present. The current heuristic is the top 100 crates by all-time downloads. We've also toyed around with the idea of having "sponsored crates", where someone contributes money towards server costs (which are paid out of our own pocket at the moment), but strangely no one seems excited to pay for it ;-) If you are, I'm sure we can get in contact.

We do [have an issue](./101) to investigate using something that would take recent popular crates into account. If Rayon would occur in that list, one way of seeing it appear on the playground would be to contribute the effort needed to add that to crates.io.

As a specific point, the playground runs on a single core (again, because we are playing out of pocket), so you'd never get to see the benefits of Rayon in runtime.

I'm going to close this as a duplicate of #101, as that seems like the most likely path that Rayon would occur, but thank you for your suggestion!

PR #157, opened 2017-07-08
Fixing build in CI, removing fingerprint munging, updating crates

shepmaster:


lilianmoraru:
Would be nice to have `rayon` in that list.
In general, would be nice to have all the crates from the Cookbook(users could experiment, maybe even the cookbook could use this playground to run the code): https://brson.github.io/rust-cookbook/

shepmaster:
@lilianmoraru please read:

1. https://github.com/integer32llc/rust-playground#whats-different
1. https://github.com/integer32llc/rust-playground/blob/master/.github/ISSUE_TEMPLATE.md
1. https://github.com/integer32llc/rust-playground/issues/113
1. https://github.com/integer32llc/rust-playground/issues/101

lilianmoraru:
I don't think that it's good that the rules are black and white.
If a crypto, concurrency, database, etc... library is not used as often as a `libc` crate(or String, Vec, File ops crate, something more basic), it doesn't mean it is less important or useful, it is just that by their nature, those crates will always be more often used(the minimal/basic ones).

The Cookbook crates seem like a good point to make an exception(it is a bit more opinionated than the first 100 crates but still no favorites picking, almost).
As for `rayon`, the idea is not to benchmark code inside the playground, it is just to be able to learn and play with the API(ex: playing with Futures + Rayon).

You made a good financial point so I won't try to "insist" further.

Issue #162, opened 2017-07-12
Have an option to split the window vertically instead of horizontally


nagisa:
In the world where 99.9% of the screens are already thread short (think 16:9, 21:9, 23:9, evenmore:9), vertical space is at super premium and yet the result screen is under a horizontal split.

Given the fact that on a standard 1080p monitor the 80 symbol delimiter is way off to the “left” from the centre, maybe it would be a great idea to use a vertical split instead as the old playground used to be?

shepmaster:
The old playground actually also split horizontally, depending on how wide the window was. I never noticed because my resolution wasn't wide enough to trigger it, but other people brought it up. I do have a branch around here somewhere with a partial implementation.

denis-golubev:
I'd like to help with this and have already managed to add a configuration option in the configuration dialog (or should this be a button in the main ui?). Based on this configuration the flex direction of the Editor- and Output-divs should be changed (currently a new div which contains both of them just collapses to zero height). I will create a pull request later today, but this is what I've done so far: https://github.com/denis-golubev/rust-playground/commit/9ad1b468f741fdcb8dacea8681117bd087f12db6

Another question is the layout of the tabs of the output pane if the view is split vertically. I think they should still be in a row, when the output is shown, but in a column, when hidden.

shepmaster:
@denis-golubev that's great! 

> a configuration option in the configuration dialog (or should this be a button in the main ui?)

A configuration option sounds perfect.

One thing that's been discussed elsewhere is actually having 3 states: "automatic", "horizontal", "vertical". The latter two map to what you have added, "automatic" would be closer to what the old playground did - adding a media query based on the size of the screen:

```css
@media screen and (max-width: 1200px) { ... }
```

> the layout of the tabs of the output pane if the view is split vertically

Hmm, I hadn't even thought about that! It does look a little weird, doesn't it?

![image](https://user-images.githubusercontent.com/174509/28167240-e6d46fa4-67a8-11e7-9afd-6f957e1dc0ba.png)

I think that it would be OK to leave them as a horizontal row to start with. I'm sure the people clamoring for a side-by-side view would rather have it sooner rather than later. It's also possible that the styling for those tabs may or may not be easy to adjust for changing orientation... I vaguely remember there being some ugliness around that CSS.

----

One stylistic issue that my code had and you may want to look out for, is the lack of spacing between the side-by-side panes and extra spacing on the top:

[...]

PR #174, opened 2017-07-13
Add configurable split behaviour

denis-golubev:
As discussed in issue #162, there is now an automatic option, too. I tried to fix the the margins, but the current CSS solution is hacky and I think this should be revisited. Perhaps when adding the rotation for the hidden tab buttons in vertical split mode? I just added the split configuration for now because it has higher priority.

I'll be happy about feedback :)

### Description:

Add an option in the configuration menu to change the split behaviour
/ orientation between "Automatic" (similar to the old playground),
"Horizontal" (the current behaviour of the integer32llc playground) and
"Vertical" (similar to the old playground for widescreen displays)

### Detailed:
- Add changeOrientation and orientation to the props of Configuration
- Add a new ConfigurationSelect-instance in the Configuration
componenent
- Add said orientation state to the props of Playground
- Introduce new 'div' which contains the Editor- and the Output-divs
and has a CSS class depending on the current orientation setting
- Add a new changeOrientation-action

### CSS changes:
- Add two @mixin which are active, when during vertical split mode and
automatic split mode with a width of at least 1600px
- Add a hacky solution for margin-top of the output pane

### DIFF
--- a/ui/frontend/Configuration.jsx
+++ b/ui/frontend/Configuration.jsx
@@ -9,6 +9,7 @@ import {
   changeEditor,
   changeKeybinding,
   changeTheme,
+  changeOrientation,
   toggleConfiguration,
 } from './actions';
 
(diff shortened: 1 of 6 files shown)


shepmaster:
> but the current CSS solution is hacky and I think this should be revisited

That's a polite understatement. Since I first wrote the CSS for the playground, I've come to learn and love [BEM](http://getbem.com/), so I plan on cleaning that up a bit at some point.

denis-golubev:
@shepmaster I'm sorry! I was just talking about my solution!

shepmaster:
Haha, no worries. I didn't think that you were being mean ;-) Just looking through the code while reviewing made me realize how much I've learned in the last year.

I've merged and the automatic deploy should have it out on the real world in an hour or so.

Thanks again!

Issue #135, opened 2017-04-24
Load some/all of the files in a gist if the specific file name is not found


shepmaster:
Works:

https://play.rust-lang.org/?gist=f6a1e1ec29da1ee73994820f742cdb7d&version=stable&backtrace=0

No works:

https://play.integer32.com/?gist=f6a1e1ec29da1ee73994820f742cdb7d&version=stable&backtrace=0

PR #208, opened 2017-09-02
Load code from gist.github.com with any filename

dns2utf8:
Hi all

I noticed that all my gist links (eg. [this](https://play.rust-lang.org/?gist=828d338bf3cb4214a8127786d89fe085&version=stable) or [that](https://play.rust-lang.org/?gist=363d161c570fef39b555f6e08c8f7d00&version=stable&backtrace=2)) were broken because I use a more specific filename than `playground.rs`.

With this fix the UI will load the last/only file inside a gist so any name can be chosen.

Best

### DIFF
--- a/ui/frontend/gist.js
+++ b/ui/frontend/gist.js
@@ -14,11 +14,27 @@ const FILENAME = 'playground.rs';
 export function load(id) {
   return fetch(`${baseUrlStr}/${id}`)
     .then(response => response.json())
-    .then(gist => ({
-      id: id,
-      url: gist.html_url,
-      code: gist.files[FILENAME].content,
-    }));
+    .then(gist => {
+      const filenames = Object.keys(gist.files);
+      if (filenames.length > 0) {
+        let code = gist.files[filenames[0]].content;
+        if (filenames.length > 1) {
+          code = filenames.reduce(
+                            (code, filename) => `${code}\n\n// ${filename}\n\n${gist.files[filename].content}`
+                            , '')
+                            .trimLeft();
+        }
+        return {
+          id: id,
+          url: gist.html_url,
+          code: code,
+        };
+      } else {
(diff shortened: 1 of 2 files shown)


ui/frontend/gist.js
```diff
@@ -14,11 +14,17 @@ const FILENAME = 'playground.rs';
...
+    .then(gist => {
+        let filename = FILENAME;
+        for (let x in gist.files) {
+            filename = x;
+        }
+        return {
+            id: id,
+            url: gist.html_url,
+            code: gist.files[filename].content,
+        };
```
shepmaster: > With this fix the UI will load the last/only file inside a gist so any name can be chosen.

How about we go ahead and change the code to reflect that reality? In that case, we could just use 

```js
code: gist.files.pop().content,
```

without doing the extra looping through the files and the unused initial setting of the file.

You should also address the linting issues.

---

If you wanted to be really advanced and fully solve #135...

I'd check to see how many files there are. If there's only one, just use that. If there are multiple, I'd combine them all, adding a comment before each one with the name of the file.

dns2utf8:
I have never used the old playpen with more than one file so I can not compare it.
Howerver I implemented it more functional.

Sadly I am not familiar with the futures in JavaScript. Is there a good way to return an error?

shepmaster:
Thanks! The change looks good. If you'd squash all the commits, I'll merge it. 

> I have never used the old playpen with more than one file so I can not compare it.
> Howerver I implemented it more functional.

I don't know *if* the old playground actually supported multiple files :-)

> Sadly I am not familiar with the futures in JavaScript. Is there a good way to return an error?

Overall, the playground doesn't have great error reporting right now. As you can see, the old code assumed there was a file of a specific name and just failed when it was missing. This is at least as good as that.

dns2utf8:
Cool 😄  Squashed them.

shepmaster:
Closes #135

PR #213, opened 2017-09-16
Rename "Gist" button to "Share"

panicbit:


shepmaster:
Seems reasonable to me, thanks!

Issue #222, opened 2017-10-30
layout problem on config


mone27:
When you open the config  menu there are some layout problem: the "done" button is placed over a menu and the last entry on the menu is outside the white box.
![screenshot from 2017-10-30 10-45-59](https://user-images.githubusercontent.com/17617810/32164908-2acb94ea-bd61-11e7-80ac-5f51f5731be3.png)
I have Ubuntu 17.10 and the bug is present on Firefox Nightly 58.0a, Firefox 56.0, Chromium 62 as well as Epiphany 3.26.1.(all the browser I have tested)
On the other hand on mobile(android 8.0) the bug isn't present neither on Firefox Nightly nor Chrome.
I hope this report would be useful.
I'm willing to help fixing it if I can understand better where it is

shepmaster:
Very strange! I'm on macOS 10.12.6 and don't see that:

**Firefox 56.0**

![image](https://user-images.githubusercontent.com/174509/32171683-6714f5e8-bd4f-11e7-8569-03390568bffd.png)

**Chrome canary 64.0.3253.0**

![image](https://user-images.githubusercontent.com/174509/32171718-82c249e4-bd4f-11e7-9577-ee71d332dc44.png)

**Safari 11.0 (12604.1.38.1.7)**

![image](https://user-images.githubusercontent.com/174509/32171758-9fca3d8a-bd4f-11e7-80dd-40a343a8fa97.png)

shepmaster:
You may need to do more investigation to narrow down the root cause — isn't CSS fun! Otherwise, I'll need to wrangle up a Ubuntu VM, which will probably take a while.

mone27:
I've  tested it on other platforms and it seems a Linux issue.
It works on windows (both Firefox 56 and edge)
I tested it on Ubuntu 16.04 lts (Firefox 56) and on kali Linux (Firefox esr 52) and it doesn't work on any of those systems. 
It remains to test it under KDE because could be gtk issue. 
I'll dive into CSS during next days

hcpl:
Can confirm this under Ubuntu 14.04 (i3wm, Firefox 56):

![2017-10-30-195525-646x258_maim_slop](https://user-images.githubusercontent.com/17436405/32187031-711d51cc-bdac-11e7-88c0-60f1f49c5180.png)

shepmaster:
This is "fixed by accident" by... deleting the config modal! :-) /cc #299

mone27:
[...]

PR #228, opened 2017-11-15
Add demangling and directive hiding for asm

chandlore:
Implements symbol demangling and the option to hide assembler directives (e.g. ".loc 2 338 0 prologue_end") from the generated assembly.  Both options default to on and are editable using the Config dialog.

**New - Demangled with Directives Removed**:
```asm
playground::main:
	sub	rsp, 56
	lea	rax, [rip + ref.2]
```

**Existing - Mangled with Directives**:
```asm
	.text
	.intel_syntax noprefix
	.file	"playground.cgu-0.rs"
	.section	.text._ZN10playground4main17hf4c1d0e5656172f7E,"ax",@progbits
	.p2align	4, 0x90
	.type	_ZN10playground4main17hf4c1d0e5656172f7E,@function
_ZN10playground4main17hf4c1d0e5656172f7E:
	.cfi_startproc
	sub	rsp, 56
.Lcfi0:
	.cfi_def_cfa_offset 64
	lea	rax, [rip + ref.2]
```


ui/frontend/index.scss
```diff
@@ -427,7 +427,7 @@ $PHI: 2.61803398875;
         margin-top: 10em;
 
         width: 600px;
-        height: 600px / $PHI;
+        height: 950px / $PHI;
```
shepmaster: Why is this needed?

ui/src/sandbox.rs
```diff
@@ -542,6 +567,7 @@ mod test {
     fn basic_functionality() {
         let req = ExecuteRequest {
             channel: Channel::Stable,
+            crate_type: CrateType::Binary,
```
shepmaster: Oops, I guess I don't run these tests in CI at all! I should fix that...

ui/frontend/index.scss
chandlore: I found that the background to the config popup would not fit the new rows, causing the "Done" button to be overlaid on top of the drop-downs as in issue #222.  I should probably have removed the divide by $PHI as the golden ratio is pretty well shot by this change.

chandlore: No issues with Intel vs AT&T, directives have the same format in both styles.  The [remove_assembler_directives](https://github.com/chandlore/rust-playground/blob/ec3518ca5887b75fc5ba2cb8d6982ac028eb3a20/ui/src/asm_cleanup.rs#L5-L19) function just searches for any lines that start with a period and excludes them from the updated string.

ui/src/sandbox.rs
chandlore: On my local test sandbox::test::memory_usage_is_limited is failing with:

```rust
---- sandbox::test::memory_usage_is_limited stdout ----
thread 'sandbox::test::memory_usage_is_limited' panicked at 'assertion failed: resp.stderr.contains("Killed")', sandbox.rs:848:8
```
Not sure if that's a real issue, or just differences between the containers on my machine and production.

shepmaster:
Wonderful, thank you very much!

PR #230, opened 2017-11-17
Fix labels being removed from asm

chandlore:
The assembler directive screening added in 9c495af2ce660c2253e4ea97562f52559cf5a93f is removing labels as well as directives from the assembly.  As labels are needed to understand code flow, this is bad.  This change prevents labels from being removed when assembler directives are screened out. 

A future enhancement I'd like to try is to only display labels that are used, which is the default on Godbolt, but that is a more complex change and it's better to resolve this issue quickly.

**New - Labels Retained**
```asm
.LBB0_2:
	movl	$0, -48(%rbp)
	jmp	.LBB0_4    ;Target of jmp is defined
.LBB0_4:
	movq	-48(%rbp), %rax
```
**Existing - Labels Removed**
```asm
	movl	$0, -48(%rbp)
	jmp	.LBB0_4     ;No definition of .LBB0_4, unclear where jmp goes
	movq	-48(%rbp), %rax
```

### DIFF
--- a/ui/src/asm_cleanup.rs
+++ b/ui/src/asm_cleanup.rs
@@ -4,7 +4,7 @@ use rustc_demangle::demangle;
 
 pub fn remove_assembler_directives(block: &str) -> String {
     lazy_static! {
-        static ref ASM_DIR_REGEX: Regex = Regex::new(r"(?m)^\s*\..*$").expect("Failed to create ASM_DIR_REGEX");
+        static ref ASM_DIR_REGEX: Regex = Regex::new(r"(?m)^\s*\..*[^:]$").expect("Failed to create ASM_DIR_REGEX");
     }
 
     let mut filtered_asm = String::new();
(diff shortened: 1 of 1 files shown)


shepmaster:
It looks like you've got some strange cruft commits in the PR. Would you mind rebasing on top of master? I'd expect there to only be one commit in the PR. 

If not, let me know and I can do it otherwise.

Thanks!

chandlore:
Sorry about that, please go ahead with the rebase from your end. I am a newbie with git and don't want to mess things up even more.

shepmaster:
> please go ahead with the rebase from your end

I've done so now, just waiting for the tests to be green again.

> I am a newbie with git and don't want to mess things up even more.

I'd highly recommend playing around with `git rebase`, it's a very powerful tool!

chandlore:
Thanks for your help!  I was getting "noop" as the only item displayed in the rebase flow, maybe because I had already merged the branch into my master?

BTW it was great meeting you at Rust Belt, and you folks did a great job with the conference.  I was the guy you chatted with before Chris Krycho's talk on Friday, we discussed mainframes.

shepmaster:
> Thanks for your help! I was getting "noop" as the only item displayed in the rebase flow, maybe because I had already merged the branch into my master?

That would likely because you were rebasing on top of your "upstream" branch, which was likely to be the same branch you were already on. Rebasing on top of yourself doesn't have much to do 😉. In comparison, I did something like `git rebase -i origin/master` and then removed the extra commits, leaving just the last one.

This is probably all because I squashed your previous PR into a single commit as I didn't see a need to keep the extra commits adjusting whitespace forever.

> it was great meeting you at Rust Belt, and you folks did a great job with the conference.

Thanks! It's always nice to hear that people had fun, hopefully also learning something and meeting people in the process. I do remember talking with you! I'm excited to see that you are contributing to Rust stuff... next step is contributing to the compiler, right? 😇

chandlore:
> That would likely because you were rebasing on top of your "upstream" branch, which was likely to be the same branch you were already on. Rebasing on top of yourself doesn't have much to do 😉.

Ah OK, that makes sense.  Thanks!
> next step is contributing to the compiler, right? 😇

Ha, maybe next week 😜 I will try to figure out some of the more complex filtering from Godbolt, though.

PR #235, opened 2017-11-26
Basic support for WebAssembly

### LINKED ISSUE #232: Investigate supporting WASM target
Alex suggests we need:

1. a wasm target
2. a thing that disassembles a binary wasm file to a "wast" file or "wat"

The "asm" output of llvm is not at all the "wat" format but the "wat" format is the "official" one

We should also garbage-collect the WASM to make it look better.

shepmaster:
image

Closes #232

### DIFF
--- a/compiler/base/cargo-wasm
+++ b/compiler/base/cargo-wasm
@@ -1,0 +1,51 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Rewrite our arguments to be `cargo build` instead of `cargo wasm`;
+# this assumes that the command will always be `cargo wasm ...`. We
+# capture the output directory in order to place the result file.
+shift # Ignore "wasm"
+args=()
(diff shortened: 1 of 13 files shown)


compiler/base/cargo-wasm
```diff
@@ -0,0 +1,51 @@
...
+if [[ -n "${crate_type}" ]]; then
+    cat >> Cargo.toml <<EOF
+[lib]
+crate-type = ["${crate_type}"]
+EOF
+fi
+
+# Perform our build
+# https://stackoverflow.com/a/7577209/155423
+cargo "build" "--target=wasm32-unknown-unknown" ${args[@]+"${args[@]}"}
```
notriddle: Use `"${foo[@]+"${foo[@]}"}"`. That way, it'll properly quote arguments that have spaces in them.

PR #237, opened 2017-11-29
Make WASM button disabled instead of hidden

### LINKED ISSUE #236: Feature request: On beta/stable, disable the WASM button instead of hiding it.
As `wasm-unknown-unknown` is only supported on nightly, the WASM button will be hidden when the beta/stable channel is selected. However, showing/hiding the button shifts the entire toolbar horizontally, which is annoying to use.

It is better just disable the WASM button outside of nightly, so that the toolbar won't move.

Enet4:
Resolves #236 

- make WASM button always visible, add tooltip when WASM compilation is not available
- adjust styles for disabled buttons

### DIFF
--- a/ui/frontend/Header.tsx
+++ b/ui/frontend/Header.tsx
@@ -85,11 +85,6 @@ class Header extends React.PureComponent<HeaderProps> {
 
     const primaryLabel = executionLabel(crateType, tests);
 
-    const mirButton = (
-      <button className="header-set__btn"
-        onClick={compileToWasm}>WASM</button>
-    );
-
     return (
       <div className="header">
         <div className="header-compilation header-set">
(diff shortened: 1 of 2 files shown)


ui/frontend/Header.tsx
```diff
@@ -102,7 +97,10 @@ class Header extends React.PureComponent<HeaderProps> {
               onClick={compileToLLVM}>LLVM IR</button>
             <button className="header-set__btn"
               onClick={compileToMir}>MIR</button>
-            {wasmAvailable && mirButton}
+            <button className="header-set__btn"
+              disabled={!wasmAvailable}
+              title={!wasmAvailable && 'Compilation to WASM requires a Nightly channel'}
```
shepmaster: minor nit: "the nightly"

ui/frontend/index.scss
```diff
@@ -195,18 +195,25 @@ body {
         border-left: $border;
         background: none;
 
-        cursor: pointer;
+        &:enabled {
```
shepmaster: I already had this, but why did I... buttons already set the cursor. Will have to look into that in the future.

shepmaster:
Thanks!

PR #253, opened 2018-02-02
Added filter for used labels and data directives in assembly

chandlore:
This adds more thorough filtering and formatting to the assembly view, removing unused labels and keeping relevant data directives.

The "Assembler Directives" option in the Configuration menu is renamed to a more generic "Assembly View," as more than just directives are being removed now.  Internally, "hideAssemblerDirectives" is changed to "processAssembly" for the same reason.

**New**
```asm
playground::main:
	subq	$56, %rsp
	leaq	ref.2(%rip), %rax
	movq	%rax, 8(%rsp)

str.1:
	.ascii	"Hello, world!\n"

ref.2:
	.quad	str.1
	.quad	14
```

**Current**
```asm
playground::main:
	subq	$56, %rsp
.Lcfi0:
	leaq	ref.2(%rip), %rax
	movq	%rax, 8(%rsp)
.Lfunc_end1:

str.1:

ref.2:
```

### DIFF
--- a/ui/frontend/Configuration.tsx
+++ b/ui/frontend/Configuration.tsx
@@ -7,14 +7,14 @@ import {
   changeAssemblyFlavor,
   changeDemangleAssembly,
   changeEditor,
-  changeHideAssemblerDirectives,
   changeKeybinding,
   changeOrientation,
+  changeProcessAssembly,
   changeTheme,
   toggleConfiguration,
 } from './actions';
 import State from './state';
-import { AssemblyFlavor, DemangleAssembly, Editor, HideAssemblerDirectives, Orientation } from './types';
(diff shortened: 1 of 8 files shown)


shepmaster:
> Internally, "hideAssemblerDirectives" is changed to "processAssembly" for the same reason.

As a pedantic side note, this is *technically* a kind of "breaking change", as anyone who had set this value to a non-default will find it "reset" to the new default value. I think the old setting will also take up (a tiny amount of) space in the local storage more-or-less forever.

I'm not actually worried about this case, but we do theoretically have a version number on our storage and *could* write migrations for it, if we ever have a big enough change.

shepmaster:
Thanks!

The work you are doing here seems pretty useful; you may want to consider creating a crate with the logic so it can be reused!

chandlore:
Thanks, that hadn't occurred to me.  I'll have to think about doing that.

PR #258, opened 2018-02-10
Specify codegen-units = 1 in Cargo.toml

### LINKED ISSUE #252: Set codegen units to 1
@eddyb reports that in nightly Rust, the LLVM IR output is essentially useless when multiple codegen units are in play.

For example, this doesn't always include the main function(s):

```rust
fn main() { vec![1, 2, 3]; }
```

Unfortunately, this may not be deterministic.

To address this, we can set

`codegen-units = 1` in all the appropriate profile sections in the Cargo.toml (`[profile.dev]` and `[profile.release]`, at least).

chandlore:
This fixes #252 with the suggested solution.

However, I am concerned this may not fully resolve the issue as I am still seeing the warning below after the fix:

```
warning: ignoring emit path because multiple .ll files were produced
```
According to the comment [here](https://github.com/rust-lang/rust/blob/29c8276cee4a0eab7e0634ff25c6b47bd9f87c6c/src/librustc_trans/back/write.rs#L1069-L1073), this message will display if there are multiple codegen units.

```rust
            if crate_output.outputs.contains_key(&output_type) {
                // 2) Multiple codegen units, with `--emit foo=some_name`.  We have
                //    no good solution for this case, so warn the user.
                sess.warn(&format!("ignoring emit path because multiple .{} files \
                                    were produced", ext));
```

Interestingly, if you set codegen-units > 1 in args when emitting LLVM IR and specifying the output path, the compiler will warn that this is not supported and [default codegen-units back to 1](https://github.com/rust-lang/rust/blob/afa8acce251cda7ab1548640fdb769139a45f839/src/librustc/session/config.rs#L1740-L1768).
```rust
        match codegen_units {
            Some(n) if n > 1 => {
                if matches.opt_present("o") {
                    for ot in &incompatible {
                        early_warn(error_format, &format!("--emit={} with -o incompatible with \
                                                         -C codegen-units=N for N > 1",
                                                        ot));
                    }
                    early_warn(error_format, "resetting to default -C codegen-units=1");
                    codegen_units = Some(1);
                    disable_thinlto = true;
                }
            }
```


chandlore:
Possibly related to rust-lang/rust#48029

compiler/base/Cargo.toml
```diff
@@ -3,6 +3,12 @@ name = "playground"
 version = "0.0.1"
 authors = ["The Rust Playground"]
 
+[profile.dev]
```
shepmaster: This file is actually [generated by the `top-crates` script](https://github.com/integer32llc/rust-playground/blob/master/top-crates/src/main.rs#L42-L47) as part of the calculation of top N downloads and dependencies.

We will want to move this change to the creation step so it doesn't get lost every time the crates are updated.

chandlore:
I've added the profile updates to the top-crates file.  The result.Cargo.toml file generated looks correct, although no space is added between the "package" section and "profile.dev".  I've played around the field order and whether the profile section is optional, but the formatting issue remained.  Might be related to alexcrichton/toml-rs#219.

```toml
[package]
name = "playground"
version = "0.0.1"
authors = ["The Rust Playground"]
[profile.dev]
codegen-units = 1

[profile.release]
codegen-units = 1

[dependencies]
adler32 = "1.0.2"
```

shepmaster:
Nice! I went ahead and removed the `Option` from the profiles for now. Although technically that's what the file has, our interpretation of it doesn't need to be optional. I also switched from a direct rename to using the `kebab-case` renamer (all the other cases actually change the name).

Since we are touching the Cargo.toml (triggering a rebuild of the containers), I also ran the script to update those files.

I force-pushed to your branch, hope you don't mind  😇.

chandlore:
I see your point about removing the Option, the profiles always be included, so there's no benefit to adding the possibility of their being left off.

Thanks for your patience, I know you're busy and I really appreciate you taking the time to help!

PR #259, opened 2018-02-11
Disable incremental compilation to resolve issue 252

chandlore:
Apologies for the onslaught of pull requests lately, but this one should resolve the underlying problem from #252.

I found that incremental compilation will disregard the value of codegen-units, causing multiple output files to be created when using -o foo --emit=bar.  This was why the issue only occurred in nightly and beta debug, stable rustc and release builds do not use incremental compilation.

I opened rust-lang/rust#48147 to rustc, but we can mitigate the issue for the playground immediately by disabling incremental compilation via Cargo.toml.

### DIFF
--- a/top-crates/src/main.rs
+++ b/top-crates/src/main.rs
@@ -77,6 +77,7 @@ struct Modifications {
 #[serde(rename_all="kebab-case")]
 struct Profile {
     codegen_units: u32,
+    incremental: bool,
 }
 
 /// Available profile types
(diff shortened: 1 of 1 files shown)


shepmaster:
Incremental is probably not useful for a case like the playground anyway, so this seems like a better place to be anyway.

shepmaster:
> Apologies for the onslaught of pull requests lately,

Never! So long as we make forward progress, it's nice to have progress on the playground. 😺

PR #261, opened 2018-02-17
Improve logic around data directive references

chandlore:
The present assembly filter is a bit overeager with label references in data directives.  This results in labels being included in filtered results that are not referenced by an opcode, directly or indirectly.

This change adds a check to ensure that label references from data directive are only treated as used if the label the data directive is part of is used by an opcode, or referenced by a data directive that is itself used by an opcode.

This should also be the final polish for the current filtering logic, at least until I notice another bug 😄


ui/src/asm_cleanup.rs
```diff
@@ -89,9 +91,12 @@ pub fn filter_asm(block: &str) -> String {
             current_label = label_decl_cap.as_str(); 
         } else if DATA_REGEX.is_match(line) && current_label != "" { 
             line_info.push(Data(current_label));
-            // Skip the data directive, just add operands
+            // These will be checked for references to other labels later on
+            // Skip the data type, just capture its reference
             for label_ref_cap in LABEL_REF_REGEX.captures_iter(line).skip(1).filter_map(|cap| cap.get(1)) {
-                opcode_operands.insert(label_ref_cap.as_str());
+                // Capture unique labels referenced by data dirs in the current label
+                let label_set = data_dir_label_refs.entry(current_label).or_insert(HashSet::new());
```
shepmaster: `.or_insert_with(HashSet::new)`

ui/src/asm_cleanup.rs
```diff
@@ -89,9 +91,12 @@ pub fn filter_asm(block: &str) -> String {
...
+                &label_set.insert(label_ref_cap.as_str());
```
shepmaster: Why is a reference being taken here?

ui/src/asm_cleanup.rs
```diff
@@ -107,7 +112,31 @@ pub fn filter_asm(block: &str) -> String {
...
+    // We only include labels ref'd by data dirs if the data dir is part of a used label
+    // Check all labels that include data dirs
+    // If label is used, treat all referenced labels from its data dirs as used
+    // Repeat this to check for any addt'l labels found until no more are located
+    let mut new_used_label_found = true;
+    let mut checked_labels: HashSet<_> = HashSet::new();
+    let mut iteration_count = 0; 
+    const MAX_ITERATIONS: u32 = 100; // Safeguard against hangs, should not be an issue but better to be safe
+
+    while new_used_label_found && iteration_count < MAX_ITERATIONS {
```
shepmaster: Since the iteration limit is a safety concern, let's promote it to the primary loop condition and make it less prone to failure. Untested changes:

```rust
    let mut checked_labels = HashSet::new();
    const MAX_ITERATIONS: u32 = 100; // Safeguard against hangs, should not be an issue but better to be safe

    for _ in 0..MAX_ITERATIONS {
        let mut new_used_label_found = false;

        for (label, references) in &data_dir_label_refs {
            if used_labels.contains(label) && checked_labels.insert(label) {
                new_used_label_found = true;

                for referenced_label in references.iter() {
                    used_labels.insert(referenced_label);
                }
            }
        }

        if !new_used_label_found {
            break;
        }
    }
```

ui/src/asm_cleanup.rs
```diff
@@ -73,7 +74,8 @@ pub fn filter_asm(block: &str) -> String {
     let mut current_label: &str = "";
     let mut line_info: Vec<LineType> = Vec::new();
     let mut labels: HashSet<&str> = HashSet::new();
-    let mut opcode_operands: HashSet<&str> = HashSet::new();
+    let mut opcode_operands: HashSet<&str> = HashSet::new(); 
+    let mut data_dir_label_refs: HashMap<&str, HashSet<&str>> = HashMap::new();
```
shepmaster: FWIW, you don't have to change this in this PR, but I'll probably go though here with Clippy and remove a bunch of the explicit types on variable declarations — they aren't very idiomatic.

ui/src/asm_cleanup.rs
```diff
@@ -73,7 +74,8 @@ pub fn filter_asm(block: &str) -> String {
     let mut current_label: &str = "";
     let mut line_info: Vec<LineType> = Vec::new();
     let mut labels: HashSet<&str> = HashSet::new();
-    let mut opcode_operands: HashSet<&str> = HashSet::new();
```
shepmaster: Why is this line changing? Is there extra whitespace?

ui/src/asm_cleanup.rs
```diff
@@ -4,6 +4,7 @@
 use regex::Regex;
 use regex::Captures;
 use rustc_demangle::demangle;
+use std::collections::HashMap;
```
shepmaster: FWIW, you don't have to change this in this PR, but I'll probably rewrite these as `std::collections::{HashMap, HashSet}` at a later point;

chandlore: I'll condense the collections and regex imports.

ui/src/asm_cleanup.rs
chandlore: Yup, must have added an extra space. Removing.

ui/src/asm_cleanup.rs
chandlore: That's a big improvement, thanks!

To be clear, I have not experienced any hangs in my testing and in practice this whole check only adds a 5% performance penalty, but I worry about someone compiling a truly massive file that interacts with this loop in ways I did not anticipate.

ui/src/asm_cleanup.rs
chandlore: Removed.

ui/src/asm_cleanup.rs
chandlore: Mistake, I misread the docs for ```or_insert```.

ui/src/asm_cleanup.rs
chandlore: What's the advantage of ```HashSet::new``` over ```HashSet::new()``` here?

shepmaster: It avoids creating the `HashSet` when the value already exists.

In *this* case, it's not a big benefit as an empty `HashSet` doesn't allocate memory and is just a few pointers-worth of stack data. In other cases, it can have strong performance implications I (and clippy!) prefer it this way to be clear that the code only executes when needed.

ui/src/asm_cleanup.rs
```diff
@@ -107,7 +110,31 @@ pub fn filter_asm(block: &str) -> String {
...
+    const MAX_ITERATIONS: u32 = 100; // Safeguard against hangs, should not be an issue but better to be safe
+
+    for _ in 0..MAX_ITERATIONS {
+        let mut new_used_label_found = false;
+
+        for (label, references) in &data_dir_label_refs {
+            if used_labels.contains(label) && checked_labels.insert(label) {
+                new_used_label_found = true;
+                
+                for referenced_label in references.iter() {
```
shepmaster: Maybe just `used_labels.extend(&references)`?

ui/src/asm_cleanup.rs
chandlore: I think I understand, the function in ```or_insert``` is eagerly evaluated, but ```or_insert_with``` is lazy, right?

If that's correct, I might put a PR in for the [entry API docs](https://doc.rust-lang.org/std/collections/hash_map/enum.Entry.html#method.or_insert) to clarify this; I had to look at the "[or" method under Result](https://doc.rust-lang.org/std/result/enum.Result.html#method.or) for the implications to be spelled out.

ui/src/asm_cleanup.rs
chandlore: Yes, that's much better.  I was bothered by the nested for loops, but wasn't coming up with anything more expressive.

I decided to just move ```references``` into ```used_labels``` instead of passing a reference, as it's not used again in the HashMap.

ui/src/asm_cleanup.rs
shepmaster: > the function in `or_insert` is eagerly evaluated, but or_insert_with is lazy, right?

That's right, except that `or_insert` doesn't take a function/closure, but only a value, so it *must* be evaluated.

ui/src/asm_cleanup.rs
```diff
@@ -107,7 +110,28 @@ pub fn filter_asm(block: &str) -> String {
...
+        for (label, references) in &data_dir_label_refs {
+            if used_labels.contains(label) && checked_labels.insert(label) {
+                new_used_label_found = true;
+                used_labels.extend(references);
+            }
+            if !new_used_label_found {
+                break;
+            }
+        }
+    }
```
shepmaster: I'm afraid I've busted your logic here... or at least I don't understand it anymore. 

On the first iteration of the `for` loop, `new_used_label_found` is false. So long as we find the current `label`, the loop will continue, but as soon as we hit the first one, the loop stops. I think it's equivalent to this:

```rust
for _ in 0..MAX_ITERATIONS {
    for (label, references) in &data_dir_label_refs {
        if used_labels.contains(label) && checked_labels.insert(label) {
            new_used_label_found = true;
            used_labels.extend(references);
        } else {
            break;
        }
    }
}
```

This seems a little strange that we'd exit the inner loop like that; maybe the if should be outside the `for`?

---

Is this algorithm essentially trying to find the reachable nodes of a graph starting from a set of nodes? If so, you might want to look at [`petgraph::GraphMap`](https://docs.rs/petgraph/0.4.11/petgraph/graphmap/struct.GraphMap.html) instead of rebuilding a graph by hand. It may even have an algorithm for finding the reachable nodes...

chandlore: Sloppy mistake, my apologies.

Yes, your summary is correct.  I'll look into petgraph, it should be an interesting to try something new.

chandlore: That was much less painful than I was expecting, thanks for the suggestion!

ui/src/asm_cleanup.rs
```diff
@@ -107,7 +110,16 @@ pub fn filter_asm(block: &str) -> String {
...
 
-    let used_labels: HashSet<_> = labels.intersection(&opcode_operands).collect();
+    let mut data_labels = Vec::new();
+    let mut used_labels: HashSet<_> = labels.intersection(&opcode_operands).collect();
+
+    // We only include labels ref'd by data dirs if directly or indirectly used by an opcode
+    used_labels.iter()
+        .filter(|label| label_graph.contains_node(label))
+        .flat_map(|n| label_graph.neighbors_directed(n, Outgoing))
+        .for_each(|node| data_labels.push(node));
```
shepmaster: Very succinct! Could this now be

```rust
    let data_labels: Vec<_> = used_labels.iter()
        .filter(|label| label_graph.contains_node(label))
        .flat_map(|n| label_graph.neighbors_directed(n, Outgoing))
        .collect();
```

I noticed the loop is gone; do we no longer need to deeply trace the graph?

chandlore: Gah, you're right!  I apologize, let me take another pass at this.

shepmaster: Thanks for putting up with all my pedantic quibbles ;-)

shepmaster: I'm guessing the loop can be broken as:

```rust
if data_labels.is_empty() { 
    break;
}
```

shepmaster: Actually, it's probably if `used_labels` doesn't change size when the `data_labels` are added...

chandlore: I don't consider "not working correctly" to be a pedantic quibble 😄 

It should be checking the full graph now, and I added a test, which I should have done earlier.

chandlore:
Thanks!  As always, thank you for all your help and patience.

Issue #192, opened 2017-07-28
Document the `playground` Cargo metadata


dns2utf8:
Hey team

I made a test with the shiny new tool: https://play.rust-lang.org/?gist=118112c0ca3014d3683dfd6db7fa99ef&version=stable

Can you enable some the features for the crate?

shepmaster:
> Can you enable some the features for the crate?

No, we do not wish to have *anything* to do with how any particular crate is exposed. Instead, you can add the `playground` feature to the libraries Cargo.toml. Here's [how Serde does it](https://github.com/serde-rs/serde/blob/26098ed877e18ce093549f9744424b3b4f83bc59/serde/Cargo.toml#L65-L68).

shepmaster:
I'm going to repurpose this issue for adding documentation about this feature.

shepmaster:
From #326

> This is based on the custom metadata configurable for docs.rs -- see https://docs.rs/about. Ours looks like:
>
> 
> ```rust 
> # Cargo.toml
> [package]
> name = "test"
> 
> [package.metadata.playground]
> default-features = true
> features = ["std", "extra-traits"]
> all-features = false
> ```

PR #267, opened 2018-02-25
Include update with playground enabled features

dns2utf8:
Currently blocked by https://github.com/uuid-rs/uuid/pull/155
This will address the original issue of #192


shepmaster:
Thanks! However, I don't really want to accept one-off updates for a single crate. Instead, these types of changes just get rolled into periodic updates of all the crates. Once the UUID crate accepts the change, releases, then our update occurs, it will be magically available!

Issue #218, opened 2017-10-20
Anonymous gists may be retired


kivikakk:
Hi there! I'm a Rustacean who is also a GitHubber, and it looks like we may be retiring anonymous gists in the future. Given the Playground's current reliance on saving to anonymous gists, I thought it'd be good to give a heads up here. (One quickly implementable workaround that comes to mind is creating an account for the Playground for its gists.)

Maybe related to #166 and some proposals there.

shepmaster:
Thank you for the heads-up! Is there somewhere where I can find more information about this? What does "retiring" entail — will all anonymous gists be disappearing? Will old ones still exist but new ones can no longer be created? One worry about a created account is rate limiting; I don't know the current rate of sharing, but I'd be worried about hitting the limits I'm aware of.

kivikakk:
The decision is still in the works (it's currently a massive spam vector), but once it becomes officially we'll likely make a blog post with a timeline. Nothing is finalised yet, but I'm fairly sure we'll keep old ones around and simply stop new ones.

We can definitely work out something around rate limiting, and will likely look to make a general solution here too.

kivikakk:
A little bump on this; it's nearing completion. (There'll almost certainly be an announcement with some lead time given, but wanted to keep everyone updated here seeing as I opened this issue and the Playground is just so useful!)

davidtwco:
GitHub have announced [in a blog post](https://github.com/blog/2503-deprecation-notice-removing-anonymous-gist-creation).

shepmaster:
@kivikakk Is there anything in particular I need to do / someone to talk to about setting up accounts for the playground? There are two primary instances of the playground (https://play.rust-lang.org/ and https://play.integer32.com/); presumably each should have an account under their respective organizations. Is that even a thing; an "organization account"?

kivikakk:
There's no "user" account for an organization which can do things like authenticate; you'll need an OAuth token for a user with the `gist` scope; see https://developer.github.com/v3/gists/#authentication. What that user is/where it lives is up to you!

shepmaster:
I've merged the fixes as part of #270, but have only deployed to play.integer32.com for right now to shake out any issues. I'll be deploying to play.rust-lang.org sometime this week (before the 19th!).

@kivikakk Is there any way I can see the user's gist rate *before* hitting the limit, so I could request an appropriate increase?

shepmaster:
All deployed now, so... onwards!

PR #270, opened 2018-03-06
Switch to server-side gists

shepmaster:
As discussed in #218, anonymous gists are going away. This means we need to have a GitHub token to create gists with, and any such token needs to be hidden from the client.

This PR attempts to keep the exact same logic previously done on the client. This includes some "forgotten" features like the lack of handling errors.


ui/src/gist.rs
```diff
@@ -0,0 +1,93 @@
...
+    let mut core = core();
+    let github = github(token, &core.handle());
+
+    let file = Content {
+        filename: None,
+        content: code,
+    };
+
+    let mut files = HashMap::new();
+    files.insert(FILENAME.into(), file);
```
carols10cents: Hm, I'm probably missing some context around how gists are represented here, in hubcaps, and on github, but it seems a little strange that you have a filename, and are using it here on line 54, but on line 49 the `hubcaps::gists::Content` gets `filename: None`?

ui/src/gist.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let options = GistOptions {
+        description: Some(DESCRIPTION.into()),
+        public: Some(true),
+        files,
+    };
+
+    let creation = github.gists().create(&options);
+
+    let gist = core.run(creation).expect("Unable to create gist");
```
carols10cents: I would add a TODO comment here like the one on line 75 for eventual better error handling of being unable to create a gist

ui/src/main.rs
```diff
@@ -185,6 +221,27 @@ fn meta_version_nightly(_req: &mut Request) -> IronResult<Response> {
...
+
+fn meta_gist_get(req: &mut Request) -> IronResult<Response> {
+    match req.extensions.get::<Router>().unwrap().find("id") {
+        Some(id) => {
+            let token = req.extensions.get::<GhToken>().unwrap().0.as_ref().clone();
+            let gist = gist::load(token, id);
+            serialize_to_response(Ok(MetaGistResponse::from(gist)))
+        }
+        None => {
+            Ok(Response::with((status::UnprocessableEntity, )))
```
carols10cents: do you need this trailing comma here?

PR #271, opened 2018-03-06
Small wasm tweaks

alexcrichton:
* Install wasm-gc from crates.io (it's published now!)
* Turn on LTO in release builds for wasm to make it extra small


shepmaster:
Thanks!

Issue #260, opened 2018-02-12
Enable edition selection


shepmaster:
Will be activated via `epoch = 2018` in Cargo.toml

Since future compilers will support multiple epochs, this makes sense to have as a UI item, instead of just a temporary `#![feature]` flag.

My first thought is to add a new configuration option:
- `default`,
- `2018`,
- `2015`

Selecting `2018` will need to only allow (AND automatically select) nightly to start with, then trickle to beta / stable. Until then, default is `2015`.

This will require that the playground gain the ability to modify the Cargo.toml. We already hack it up for WASM, so that should be cleaned up too.

Manishearth:
We can also pass this down via RUSTFLAGS

shepmaster:
Note to self: URLs loading code need to default to 2015, loading the fresh playground should use default. The value needs to be preserved in local storage as well.

Issue #211, opened 2017-09-09
Allow using panic=abort on the playground


gnzlbg:
I'd like to be able to set `panic=abort` on the playground when exploring the generated assembly of some snippets of code.

PR #272, opened 2018-03-09
Add "Compiler Options" input field to Playground header

chandlore:
This request creates an input field titled "Compiler Options" in the header section of the playground, allowing users to enter one or more -C or -Z rustc flags.  The options requested in #260 and #211 can be enabled using the new field.

image

To mitigate the risk of passing user input to the command line, the flags are run through a regex that disregards any entries not in the expected format of ```-C/-Z flag=val```.  This could be confusing for someone unknowingly using an incorrect format, as invalid entries are silently ignored.

Flags are only passed on compile requests.

### DIFF
--- a/ui/frontend/Header.tsx
+++ b/ui/frontend/Header.tsx
@@ -5,6 +5,7 @@ import Link from './uss-router/Link';
 import {
   changeChannel,
   changeMode,
+  editCompilerFlags,
   navigateToHelp,
   performClippy,
   performCompileToAssembly,
@@ -67,6 +68,8 @@ const executionLabel = (crateType, tests) => {
 };
 
 class Header extends React.PureComponent<HeaderProps> {
+  private onChange = e => this.props.onEditCompilerFlags(e.target.value);
+
   public render() {
     const {
       execute, compileToAssembly, compileToLLVM, compileToMir, compileToWasm,
@@ -75,7 +78,7 @@ class Header extends React.PureComponent<HeaderProps> {
       crateType, tests,
       toggleConfiguration, navigateToHelp,
       stableVersion, betaVersion, nightlyVersion,
-      wasmAvailable,
+      wasmAvailable, compilerFlags, onEditCompilerFlags,
     } = this.props;
 
     const oneChannel = (value: Channel, labelText, extras) =>
@@ -138,6 +141,16 @@ class Header extends React.PureComponent<HeaderProps> {
           </div>
         </div>
 
+        <div className="header-flags header-set">
+          <legend className="header-set__title">Compiler Options</legend>
+          <div className="editor">
+            <input
+              value={this.props.compilerFlags}
+              onChange={this.onChange}
+              placeholder="-C or -Z"/>
+          </div>
+         </div>
+
         <div className="header-set">
           <div className="header-set__buttons">
             <button className="header-set__btn"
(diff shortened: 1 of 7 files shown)


oberien:
> To mitigate the risk of passing user input to the command line, the flags are run through a regex that disregards any entries not in the expected format of -C/-Z flag=val.

I don't think that this allows bash-injection, because `Command` is used. Thus, afaik, each argument is passed correctly as argument and not just put into a shell to evaluate it. The way [docker run](https://docs.docker.com/engine/reference/run/#general-form) seems to work implies to me that it's also not possible to pass any arguments to docker once `IMAGE` has been specified (which is the case when we reach compiler arguments).
Anyway, I don't know how much harm can be done passing *any* user input to the rust compiler, so I think it's better to be safe than sorry.

shepmaster:
I am not inclined to accept this PR as-is. 

The header section of the playground is *already* too complicated and I wish to redesign it to work better on smaller screen sizes (try resizing the window to a smaller width to see the current failings).

Even beyond that, the playground's primary focus (as I interpret it) is on showcasing Rust to new users and people interested in Rust. Adding this type of option to such a prominent visual location would send the wrong message to this group: you need to understand what this textbox is and what needs to be entered in it *before you can even try Rust*.

For intermediate users, they still have to maintain knowledge of what options exist and what potential values there are. Is it `-Zepoch=2015` or `-Z era=2015` or `-C edition=2015`, etc. UI-restricted widgets help prevent that. There's also the question of how these manual options interact with UI-driven options (e.g. what happens when you have optimization settings).

For advanced users, this may solve the desired end problem, but there's still questions around the intended usage. Do I want to *share* the fact that I've used a specific set of flags? This would be akin to how `mode` and `channel` are a part of the permalink.

Note that the current UI already breaks some of the ideas here; given a chance to start again, I don't know that I'd add ASM/MIR/LLVM buttons to the default UI. WASM is *also* a tricky question because that's a little bit more on the marketing side ("Look how easy it is to convert Rust to WASM"!) right now, but that may change over time... 

---

All that being said, there's some core functionality enabled here that advanced users *do* want to use. I'm not sure how to best expose that, however. I've long wanted to work with some UX designer that could provide educated feedback on these types of decisions.

The last refuge of terrible UI is to completely punt on making any decisions and force every end user to do it instead. The options panel could grow a slew of checkboxes (starting with "show extra options?", but potentially growing to "show MIR?", "show ASM?" etc.) which would be unchecked by default. A similar idea would be to have a dropdown ("basic" / "advanced") which bundles everything together. This has the drawback of lack of discovery.

It's never an easy decision.

chandlore:
You raise excellent points.  I've looked into creating an autocomplete for the flag options which should reduce some confusion, but that won't address the extra clutter and size constraints.

Given that Godbolt is built for power users, it is probably better to keep the playground simple.  Closing this request.

chandlore:
As always, thanks for your thoughtful guidance!

shepmaster:
> it is probably better to keep the playground simple

I think the main thing is to keep the "default" playground simple. As the maintainer of the Playground and a "power user", I want to have my cake and eat it too!

I wouldn't recommend deleting the branch quite yet; I'm hoping that we can think up some UI solution to the root problem!

chandlore:
Sounds good, I'll work on adding autocomplete to it in the meantime.

Issue #289, opened 2018-04-08
Adds $ when importing gist with multiple files, causes compilation error


snnsnn:
When you import a gist with multiple files via url, `$` is added at the beginning of each file, which causes compilation error since it makes the code invalid. So one has to clean the imported code manually in order to run  it.

[https://play.rust-lang.org/?gist=1f059252234606216cf8721209456287&version=stable](https://play.rust-lang.org/?gist=1f059252234606216cf8721209456287&version=stable)

Here is the error:

```
Compiling playground v0.0.1 (file:///playground)
error: expected item, found `$`
 --> src/main.rs:3:1
  |
3 | $/* 
  | ^
error: aborting due to previous error
error: Could not compile `playground`.
To learn more, run the command again with --verbose.
```
Please can you remove or comment out `$`

PR #290, opened 2018-04-08
Removes $ sign when importing gist with multiple files

snnsnn:
When you import a gist with multiple files via url, $ is added at the beginning of each file, which causes compilation error:

https://github.com/integer32llc/rust-playground/issues/289

So this change fixes the problem.


shepmaster:
Haha, that's what I get by copying JavaScript code to Rust without testing ;-)

Thanks!

PR #326, opened 2018-05-12
Implement playground-specific custom metadata to control enabled features

dtolnay:
This is based on the custom metadata configurable for docs.rs -- see https://docs.rs/about. Ours looks like:

```toml
# Cargo.toml
[package]
name = "test"

[package.metadata.playground]
default-features = true
features = ["std", "extra-traits"]
all-features = false
```

This PR includes a bump of the Cargo dependency to pick up https://github.com/rust-lang/cargo/pull/5360 which has not been included in a release yet.

### Motivation

I want the `"extra-traits"` feature of Syn so that people can debug print a syntax tree. This comes up often in IRC.

```rust
extern crate syn;

fn main() {
    println!("{:#?}", syn::parse_str::<syn::Type>("u8").unwrap());
}
```


top-crates/src/main.rs
```diff
@@ -85,6 +87,34 @@ struct Profiles {
...
+impl DependencySpec {
+    fn version(&self) -> String {
+        match *self {
+            DependencySpec::String(ref version) |
+            DependencySpec::Explicit { ref version, .. } => version.clone(),
+        }
+    }
+}
+
+fn is_true(b: &bool) -> bool {
```
shepmaster: heh

shepmaster:
As when you added the original feature support: wow, thanks!

Do you think the metadata should completely replace the `playground` feature in time?

dtolnay:
Yes I think custom metadata completely replaces our earlier scheme of looking for a `"playground"` feature. That was implemented in https://github.com/integer32llc/rust-playground/pull/100 in February 2017 and custom metadata only became available for published crates in https://github.com/rust-lang/cargo/pull/4167 in June 2017. If it had been available earlier we would have used that. (I think preserving custom metadata for published crates was blocked on upgrading Cargo from rustc-serialize to Serde because rustc-serialize cannot handle deserializing data of unspecified structure.)

For now the `uuid` crate still uses a playground feature: https://github.com/uuid-rs/uuid/blob/0.6.3/Cargo.toml#L46. If you are on board with the approach in this PR, I would land this, then send a PR to uuid to switch to custom metadata, cut a uuid release, and finally remove the playground feature here.

shepmaster:
Yeah, I agree with everything you said. I'll wait just a while to see what the UUID devs have to say; I think they added that comparatively recently.

dtolnay:
In https://github.com/uuid-rs/uuid/pull/224 it looks like they are good with it. That person reviewed the original change in https://github.com/uuid-rs/uuid/pull/189.

kinggoesgaming:
Yes, I would love this change...

Just a note: the original uuid PR for this was uuid-rs/uuid#155

PR #327, opened 2018-05-12
Remove enabling a "playground" feature

dtolnay:
The "playground" feature is superseded by Cargo.toml custom metadata from #326.

The last of the top 100 crates still using this feature is uuid. I have a PR open in https://github.com/uuid-rs/uuid/pull/224 to move them to custom metadata. I don't think we need to wait for them because the playground feature doesn't do what they want anyway. The features were enabled during transitive dependency resolution but not in the Cargo.toml generated by top-crates. For example see [this playground](https://play.rust-lang.org/?gist=40ff2a86635c432ab14c48e74372dd98) which shows uuid's `"serde"` feature not enabled despite being listed as a playground feature [here](https://github.com/uuid-rs/uuid/blob/0.6.3/Cargo.toml#L46).

### DIFF
--- a/top-crates/src/main.rs
+++ b/top-crates/src/main.rs
@@ -12,7 +12,7 @@ use std::collections::btree_map::Entry;
 use std::fs::File;
 use std::io::{Read, Write};
 
-use cargo::core::{Dependency, Package, Source, SourceId, Summary};
+use cargo::core::{Dependency, Package, Source, SourceId};
 use cargo::core::registry::PackageRegistry;
 use cargo::core::resolver::{self, Method, Resolve};
 use cargo::sources::RegistrySource;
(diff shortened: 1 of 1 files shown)


top-crates/src/main.rs
```diff
@@ -158,29 +158,6 @@ impl TopCrates {
...
-fn decide_features(summary: &Summary) -> Method<'static> {
-    // Enable `playground` feature if present.
-    if summary.features().contains_key("playground") {
-        // Would put this in a lazy_static, but its type is Vec<InternedString>
-        // and InternedString is not publicly nameable.
-        let features = Method::split_features(&["playground".to_owned()]);
-
-        Method::Required {
-            dev_deps: false,
-            features: Box::leak(Box::new(features)),
```
shepmaster: Whew, that was in there for too long!

shepmaster:
Thank you!

PR #354, opened 2018-07-09
Fix failing sandbox tests due to missing edition

chandlore:
This commit fixes errors on a number of sandbox tests. An edition was added to all compile and execute request tests; stable channel tests received a value of None, the nightly test uses Edition::Rust2018.

In addition, Clippy has changed the ['eq_op' lint](https://rust-lang-nursery.github.io/rust-clippy/current/index.html#eq_op) from warn to deny. Updated the 'linting_code' test to expect this.


shepmaster:
I *really* need to get these running in CI...

Issue #387, opened 2018-08-22
Rework Miri build process


shepmaster:
Miri has a [toolchain file](https://github.com/solson/miri/blob/master/rust-toolchain) but we want to always try to use the latest nightly.

Look into marking these builds as allowable failures in Travis as well, instead of `|| true`.

RalfJung:
> Look into marking these builds as allowable failures in Travis as well, instead of || true.

I assume this does not apply to miri only. Also, for reference:
https://docs.travis-ci.com/user/customizing-the-build/#rows-that-are-allowed-to-fail

PR #388, opened 2018-08-22
Fix miri build process

RalfJung:
However, I am not sure if I got that travis syntax right and their validator is deprecated without replacement... I guess we'll see in the PR's travis run?

Fixes https://github.com/integer32llc/rust-playground/issues/387


RalfJung:
I am seeing if the CI stuff makes any sense at https://travis-ci.org/RalfJung/rust-playground/

RalfJung:
All right, judging from https://travis-ci.org/RalfJung/rust-playground/builds/419195985 this works now.

PR #396, opened 2018-08-29
Update petgraph version to 0.4.13

mikerite:
petgraph versions before 0.4.13 no longer build on rust nightly due to a
conflict with the now stable `Iterator::find_map` method. See
https://github.com/bluss/petgraph/issues/223.


shepmaster:
Thanks!

PR #398, opened 2018-09-02
Show versions of tools.

### LINKED ISSUE #300: Show versions of Clippy / Rustfmt / Miri
Spawned from #184. 

This should probably include the *Rust* version as well, since it can be decoupled from what's in the nightly.

mikerite:
Closes #300


shepmaster:
Thanks! I made some small stylistic tweaks and squashed it all down into one commit. Once the tests pass, we'll merge!

PR #407, opened 2018-09-18
Update assembly filter to handle LLVM comments

### LINKED ISSUE #406: "Name Filtering" option is removing necessary ASM labels
I don't know if this is exactly your problem, but godbolt's ASM for this example (https://rust.godbolt.org/z/_Wczb9) doesn't have the problem, so it might be.

Repro: https://play.rust-lang.org/?gist=f5e688bd60fb8de5018b31d4f05e7ab5&version=nightly&mode=release&edition=2015

```asm
.LBB6_9:
	lea	r14, [r13 - 1]
	lea	r15, [r12 + r14]
	lea	rdi, [r12 + r13]
	mov	ebx, 1
	mov	rsi, r15
	mov	rdx, rbx
	call	memcpy@PLT
	add	r13, rbx
	mov	rbx, r13
	sub	rbx, r14
	mov	rdx, rbp
	sub	rdx, r13
	lea	rdi, [r12 + r13]
	cmp	rdx, rbx
	ja	.LBB6_10
	mov	rsi, r15
	call	memcpy@PLT
	mov	rax, qword ptr [rsp]    # 8-byte Reload
	mov	qword ptr [rax + 16], rbp
	add	rsp, 8
	pop	rbx
	pop	r12
	pop	r13
	pop	r14
	pop	r15
	pop	rbp
	ret
```
Note the jump to `LBB6_10`, which is a label not in the output.

Turn off name filtering, and you get instead
```asm
.LBB6_9:
	lea	r14, [r13 - 1]
	lea	r15, [r12 + r14]
	lea	rdi, [r12 + r13]
	mov	ebx, 1
	.p2align	4, 0x90
.LBB6_10:                               # =>This Inner Loop Header: Depth=1
	mov	rsi, r15
	mov	rdx, rbx
	call	memcpy@PLT
	add	r13, rbx
	mov	rbx, r13
	sub	rbx, r14
	mov	rdx, rbp
	sub	rdx, r13
	lea	rdi, [r12 + r13]
	cmp	rdx, rbx
	ja	.LBB6_10
# %bb.11:
	mov	rsi, r15
	call	memcpy@PLT
	mov	rax, qword ptr [rsp]    # 8-byte Reload
	mov	qword ptr [rax + 16], rbp
	add	rsp, 8
	.cfi_def_cfa_offset 56
	pop	rbx
	.cfi_def_cfa_offset 48
	pop	r12
	.cfi_def_cfa_offset 40
	pop	r13
	.cfi_def_cfa_offset 32
	pop	r14
	.cfi_def_cfa_offset 24
	pop	r15
	.cfi_def_cfa_offset 16
	pop	rbp
	.cfi_def_cfa_offset 8
	ret
```
Which has the `LBB6_10` label, as it should.

(Sorry for not reducing the repro.)

chandlore:
Fixes #406 

The LLVM comments recently added to assembly are not handled by the existing filtering scripts, causing some labels to be removed incorrectly.

This commit makes the following updates:
   * Fixes the assembly filtering to recognize labels with a comment on the same line
   * Retain comments that are on their own line
      * For now I've  retained the `# %bb` comments as seen in the example below.  I'm not sure how much utility they provide, but they also don't create too much noise.
   * Directive types that were previously removed will continue to be removed even if they have a comment
      * This removes the "Begin function..." comments added to `.p2align` directives, but still retains the standalone "End function" comments.  I don't think this will be confusing as the filtering makes it easy to tell which function we're in.  Below is an example for reference:

**Raw**
```asm
	.p2align	4, 0x90         # -- Begin function <T as core::any::Any>::get_type_id
	.type	<T as core::any::Any>::get_type_id,@function
<T as core::any::Any>::get_type_id: # @"<T as core::any::Any>::get_type_id"
	.cfi_startproc
# %bb.0:
	movabsq	$7549865886324542212, %rax # imm = 0x68C68305D036CB04
	retq
.Lfunc_end0:
	.size	<T as core::any::Any>::get_type_id, .Lfunc_end0-<T as core::any::Any>::get_type_id
	.cfi_endproc
                                        # -- End function
```

**Filtered**
```asm
<T as core::any::Any>::get_type_id: # @"<T as core::any::Any>::get_type_id"
# %bb.0:
	movabsq	$7549865886324542212, %rax # imm = 0x68C68305D036CB04
	retq
                                        # -- End function
```

### DIFF
--- a/ui/src/asm_cleanup.rs
+++ b/ui/src/asm_cleanup.rs
@@ -34,13 +34,13 @@ pub fn filter_asm(block: &str) -> String {
     lazy_static! {
         // Example:    mov rax, rdx
         // Always inlude in results
-        static ref OPCODE_REGEX: Regex = Regex::new(r"^\s*[a-zA-Z]+.*[^:]$").unwrap();
+        static ref OPCODE_REGEX: Regex = Regex::new(r"^\s+[a-zA-Z]+.*[^:]$").unwrap();
     }
     lazy_static! {
         // Example:.Lfunc_end7:
         // Finds label declarations
         // Include in results only if it is referenced by an opcode, or is a function
-        static ref LABEL_DECL_REGEX: Regex = Regex::new(r"([a-zA-Z_.<][a-zA-Z0-9$&_.,<>\[\]{}:' ]*):$").unwrap();
+        static ref LABEL_DECL_REGEX: Regex = Regex::new(r"^([a-zA-Z_.<][a-zA-Z0-9$&_.,<>\[\]{}:' ]*):(\s+#.*)?$").unwrap();
     }
     lazy_static! {
         // Example:    mov lea rdi, [rip + str.0] // str.0 is the referenced label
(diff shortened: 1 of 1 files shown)


shepmaster:
@scottmcm do you have any opinions on the improved output?

chandlore:
Interesting paper re reordering basic blocks (bb):

https://arxiv.org/abs/1809.04676

PR #410, opened 2018-09-21
Add version details for Miri

mikerite:


shepmaster:
Thanks! I poked @RalfJung to add this to Miri and then never updated our side 😊

PR #414, opened 2018-10-10
Feature: copying embedded code link

weihanglo:
Hey maintainers, this is my first contribution to rust-playground. If I did something inappropriate or violating any rule, just tell me directly. Thanks.

## Goal

`code` querystring is an awesome feature to share code instantly without generating additional gist. Hence, I implemented a proof of concept that copying URL with `code` query. Currently, this feature is placed in **Gist.tsx** component behind the **SHARE** button. If you consider this helpful, I am willing to complete it in an independent UI without creating any gist.

## Steps To Test

1. Click SHARE button
2. Click copy icon right after **Embedded code in link**.

## Caveats

Although some browsers may set a maximum length limit on their URL location bar, it seems that Firefox and Chrome support at least 10k lines of code. Here are screenshots of my manual tests on Firefox nightly 64.0a1.

1. generate 10k lines of code 

2. open URL with 10k lines embedded

### DIFF
--- a/ui/frontend/selectors/index.ts
+++ b/ui/frontend/selectors/index.ts
@@ -162,21 +162,26 @@ const baseUrlSelector = (state: State) =>
 const gistSelector = (state: State) =>
   state.output.gist;
 
+// Selects url.query of build configs.
+const urlQuerySelector = createSelector(
+  gistSelector,
+  gist => ({
+    version: gist.channel,
+    mode: gist.mode,
+    edition: gist.edition,
+  }),
+);
+
 export const showGistLoaderSelector = createSelector(
   gistSelector,
   gist => gist.requestsInProgress > 0,
 );
 
 export const permalinkSelector = createSelector(
-  baseUrlSelector, gistSelector,
-  (baseUrl, gist) => {
+  baseUrlSelector, urlQuerySelector, gistSelector,
+  (baseUrl, query, gist) => {
     const u = url.parse(baseUrl, true);
-    u.query = {
-      gist: gist.id,
-      version: gist.channel,
-      mode: gist.mode,
-      edition: gist.edition,
-    };
+    u.query = { ...query, gist: gist.id };
     return url.format(u);
   },
 );
@@ -238,6 +243,15 @@ export const issueUrlSelector = createSelector(
     const newIssueUrl = url.parse('https://github.com/rust-lang/rust/issues/new', true);
     newIssueUrl.query = { body: snippet };
     return url.format(newIssueUrl);
+  },
+);
+
+export const codeUrlSelector = createSelector(
+  baseUrlSelector, urlQuerySelector, gistSelector,
+  (baseUrl, query, gist) => {
+    const u = url.parse(baseUrl, true);
+    u.query = { ...query, code: gist.code };
+    return url.format(u);
   },
 );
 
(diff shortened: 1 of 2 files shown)


ui/frontend/selectors/index.ts
```diff
@@ -241,6 +242,16 @@ export const issueUrlSelector = createSelector(
   },
 );
 
+export const codeUrlSelector = createSelector(
+  codeSelector,
+  code => {
+    const codeUrl =  url.parse(window.location.href, true);
+    codeUrl.query.code = code;
+    codeUrl.search = qs.stringify(codeUrl.query.code);
```
shepmaster: Why are you setting both `query` and `search`? Check out the previous `*UrlSelector` functions which just set the `query` parameter and rely on the url library to properly escape the value. Then we don't need to import `qs` at all.

weihanglo: It works when no any other query presents. However, if you switch channels or alter other configs that add query strings, `url.format` would ignore your query object and use `urlObj.search` directly. See [`url.format(urlObj)`](https://www.npmjs.com/package/url/v/0.11.0#urlformaturlobj).

I know it is a little hacky here. Here is an alternative approach using `URLSearchParams` instead of `url` module. The drawback is that it does not support IE or older browsers.

```rust
code => {
    const codeUrl = new URL(window.location.href);
    codeUrl.searchParams.append('code', code);
    return codeUrl.href;
  },
```

shepmaster: > switch channels or alter other configs that add query strings

Then let's reuse the code that generates those query strings (`permalinkSelector`).

Something like:

```js
export const xxxSelector = createSelector(
  baseUrlSelector, gistSelector,
  (baseUrl, gist) => {
    const u = url.parse(baseUrl, true);
    u.query = {
      version: gist.channel,
      mode: gist.mode,
      edition: gist.edition,
    };
    return u;
  },
);

export const permalinkSelector = createSelector(
  xxxSelector, gistSelector,
  (u, gist) => {
    u.query.gist = gist.id;
    return url.format(u);
  },
);

export const codeUrlSelector = createSelector(
  xxxSelector, codeSelector,
  (u, code) => {
    u.query.code = code;
    return url.format(u);
  },
);
```

weihanglo: The example above cannot work. It is a misuse of selectors. Selectors memoize data selected and continue passing by reference. Modify argument would influence all selectors using it. 

Back to my modified code similar to this example, the **permalink** link would get a `code=fn main()...` query string.

shepmaster: > , a selector must be a pure function without any modification of arguments.

Fair enough, but the core idea is still valid:

```js
export const xxxSelector = createSelector(
  gistSelector,
  (gist) => ({
    version: gist.channel,
    mode: gist.mode,
    edition: gist.edition,
  }),
);

export const permalinkSelector = createSelector(
  baseUrlSelector, xxxSelector, gistSelector,
  (baseUrl, xxx, gist) => {
    const u = url.parse(baseUrl, true);
    u.query = { gist: gist.id, ...xxx };
    return url.format(u);
  },
);

export const codeUrlSelector = createSelector(
  baseUrlSelector, xxxSelector, codeSelector,
  (baseUrl, xxx, code) => {
    const u = url.parse(baseUrl, true);
    u.query = { code, ...xxx };
    return url.format(u);
  },
);
```

weihanglo: Modified at 3479820.

shepmaster:
After reviewing your commits, I squashed the review fixes and pushed to your branch.

weihanglo:
@shepmaster Seems great to me!

PR #415, opened 2018-10-10
Click help hint link to enable feature gate

weihanglo:
close #317

## Goal

When error message contains help hint as below,

```
error[E0658]: non-ascii idents are not fully supported. (see issue #28979)
 --> src/main.rs:4:9
  |
4 |     let α = 0.00001f64; // #![feature(non_ascii_idents)]
  |         ^
  |
  = help: add #![feature(non_ascii_idents)] to the crate attributes to enable
```

Highlight `add #![feature(non_ascii_idents)]` and provide a clickable link to insert code at  the first line to enable corresponding feature gate.

## Result

Output standard error are highlighted as following image. It seems a bit dumb. Maybe click-to-add anchors need additional styling?

screen shot 2018-10-11 at 03 49 30

## Tests

I cannot figure out how to bootstrap Ruby environment to run rspec. Therefore, this feature only be tested manually. Below is a valid test case.

```rust
const X: u8 = { let x = 1; x }; // #![feature(const_let)]

fn main() {
    let α = 0.00001f64; // #![feature(non_ascii_idents)]
}
```

PR #416, opened 2018-10-11
Remove typo single quote for adding main function

CYBAI:
We changed from string concatenation to string interpolation in https://github.com/integer32llc/rust-playground/pull/415.

With using string interpolation, we don't need the single quote now and it will cause the `fn` to be `'fn`.

| As is | To be |
| ----- | ----- |
| default | default |


shepmaster:
Oops!

Issue #22, opened 2016-07-15
Consider offering a way to specify standard input


shepmaster:

shepmaster:
One idea: Have a big ol' textbox that gets piped in to stdin.

PR #417, opened 2018-10-14
wip: Add stdin

letmutx:
Hi! This PR attempts to add stdin support(#22) for playground server. Only basic functionality is implemented. I will attempt to work on the front-end changes also if you're okay with the limitations.

Input is static; it must be sent in the execute request's payload along with the code(like [here](https://www.codechef.com/ide)) as opposed to interactive; but I am not sure whether interactive input will be useful with the current 10s execution time limit. Let me know of your thoughts. Thanks!

### DIFF
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -1173,4 +1173,31 @@ mod test {
 
         assert!(resp.stderr.contains("Cannot fork"));
     }
+
+    const STDIN_READER_CODE: &str = r#"
+        use std::io::{self, Read};
+        fn main() -> io::Result<()> {
+            let mut buffer = String::new();
+            let stdin = io::stdin();
+            let mut handle = stdin.lock();
+            handle.read_to_string(&mut buffer)?;
+            println!("{}", buffer);
+            Ok(())
+        }
+    "#;
+
+    #[test]
+    fn test_stdin_is_processed() {
+        let req = ExecuteRequest {
+            channel: Channel::Stable,
+            code: STDIN_READER_CODE.to_string(),
+            stdin: Some("something".to_owned()),
+            ..ExecuteRequest::default()
+        };
+
+        let sb = Sandbox::new().expect("Unable to create sandbox");
+        let resp = sb.execute(&req).expect("Unable to execute code");
+
+        assert!(resp.stdout == "something\n");
+    }
 }
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -5,7 +5,7 @@ use std::fs::{self, File};
 use std::io::prelude::*;
 use std::io::{self, BufReader, BufWriter, ErrorKind};
 use std::path::Path;
-use std::process::Command;
+use std::process::{Command, Stdio};
 use std::string;
 
 #[derive(Debug, Deserialize)]
@@ -59,6 +59,15 @@ quick_error! {
         UnableToReadOutput(err: io::Error) {
             description("unable to read output file")
             display("Unable to read output file: {}", err)
+            cause(err)
+        }
+        UnableToOpenStdin {
+            description("unable to open stdin")
+            display("Unable to open stdin")
+        }
+        UnableToWriteInput(err: io::Error) {
+            description("unable to write input")
+            display("Unable to read write input: {}", err)
             cause(err)
         }
         UnableToParseCrateInformation(err: ::serde_json::Error) {
@@ -167,9 +176,21 @@ impl Sandbox {
 
     pub fn execute(&self, req: &ExecuteRequest) -> Result<ExecuteResponse> {
         try!(self.write_source_code(&req.code));
-        let mut command = self.execute_command(req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req.edition);
 
-        let output = try!(command.output().map_err(Error::UnableToExecuteCompiler));
+        let mut command = self.execute_command(req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req.edition);
+        let output = if let Some(ref input) = req.stdin {
+            command.stdin(Stdio::piped());
+            command.stdout(Stdio::piped());
+            command.stderr(Stdio::piped());
+            let mut child = try!(command.spawn().map_err(Error::UnableToExecuteCompiler));
+            {
+                let stdin = try!(child.stdin.as_mut().ok_or(Error::UnableToOpenStdin));
+                try!(stdin.write_all(input.as_bytes()).map_err(Error::UnableToWriteInput));
+            }
+            try!(child.wait_with_output().map_err(Error::UnableToExecuteCompiler))
+        } else {
+            try!(command.output().map_err(Error::UnableToExecuteCompiler))
+        };
 
         Ok(ExecuteResponse {
             success: output.status.success(),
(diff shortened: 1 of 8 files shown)


ui/src/sandbox.rs
```diff
@@ -167,9 +176,21 @@ impl Sandbox {
...
         try!(self.write_source_code(&req.code));
-        let mut command = self.execute_command(req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req.edition);
 
-        let output = try!(command.output().map_err(Error::UnableToExecuteCompiler));
+        let mut command = self.execute_command(req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req.edition);
+        let output = if let Some(ref input) = req.stdin {
+            command.stdin(Stdio::piped());
+            command.stdout(Stdio::piped());
+            command.stderr(Stdio::piped());
+            let mut child = try!(command.spawn().map_err(Error::UnableToExecuteCompiler));
```
shepmaster: Feel free to use `?` instead of `try!` — this code is just comparatively old and predates `?`. I'm half-planning on doing a once-over the code to try and clean things up when the 2018 edition lands in stable.

ui/src/sandbox.rs
```diff
@@ -291,8 +312,8 @@ impl Sandbox {
 
     fn write_source_code(&self, code: &str) -> Result<()> {
         let data = code.as_bytes();
-
         let path = self.input_file.as_ref();
+
```
shepmaster: Seemingly unrelated whitespace changes

ui/src/sandbox.rs
```diff
@@ -651,6 +673,7 @@ pub struct ExecuteRequest {
     pub tests: bool,
     pub backtrace: bool,
     pub code: String,
+    pub stdin: Option<String>,
```
shepmaster: Would it make sense to drop the `Option` — is there any conceptual difference between passing in an empty string?

shepmaster:
> Input is static; it must be sent in the execute request's payload along with the code [...] as opposed to interactive

I think this is a fine compromise. Doing anything interactive would require a relatively large restructuring and rethinking of how the entire playground is architected.

ui/src/sandbox.rs
letmutx: I just wanted to keep it inline with the style used everywhere.

ui/src/sandbox.rs
letmutx: Shouldn't make much of a difference other than the fact that `stdin` is optional in the request payload.

crumblingstatue:
> Input is static; it must be sent in the execute request's payload along with the code [...] as opposed to interactive

This is also the behavior of [ideone](https://ideone.com/), which I love for the `stdin` functionality. I wouldn't even expect an online tool like this to have interactive stdin. They are for demonstration purposes, not to build interactive apps.

ui/src/sandbox.rs
shepmaster: It's not everywhere though 😉 Newer code, like the tests, use `?`. Anyway, use whichever you prefer and it will get cleaned up at some point.

letmutx:
@shepmaster Do you have any preferences on how the input box should look/positioned? I have a minimal setup working now.([picture](https://user-images.githubusercontent.com/26674948/47104808-e88bc400-d25f-11e8-8593-5fca8fe6ce5c.png)) My idea is to have an always available Input button at the bottom(see the `Execution` button in the picture) which would open up the text box when clicked just like an output pane.

lqd:
First of all, nice job on the PR! 👍 

I had a few thoughts and questions and wanted to check with you and @shepmaster:
- what kinds of inputs are we expecting to support here ? a word, a line, a paragraph ? 
- and some questions regarding usage probability/popularity: how often will it be needed, how visible/accessible does it need to be, was this feature requested often ?

The first question will help with picking a UI component for it. If it's small like a word or a short-ish line, you could imagine tucking it away in the "..." advanced options popover. If it's bigger than that, it'd have to be a textarea-like multiline input as you've done. This could be modeled as the other "panels" (Execution, Miri, Share, etc). One could maybe imagine uploading an input file, I don't know ? (I have never used a custom input in a playground-like environment and am unfamiliar with the use cases for a non-interactive session).

The second question(s) will help with finding a location for the component if it's very likely to be used and needing to be accessible at all times, and/or a way to trigger it appearing if it's a less probable use-case. In the latter cases, apart from having an input in the "..." popover, one could also have a "Custom input" button or checkbox there (or in the Config popover) to make the input area appear.

It feels to me (but without data to back it up) that this is a use case that is less common than what the playground offers today, but still has value. It also seems to me that a single word or line wouldn't that useful, vs say a couple lines or a short paragraph. That makes me think having the input area constantly visible (or a dedicated button as in the screenshot, maybe replacing the "Execution" tab in that screenshot as mentioned in the comment) may be a detriment to all users (using precious vertical space on mobile or narrow screens, etc), but also that a simple input as I suggested above wouldn't cut it either.

So what do you @letmutx and @shepmaster feel about a toggle in the "..." popover titled "Custom input", similar to the ones we have today — basically the "Backtrace enabled/disabled" one. When enabled, it would open a multiline input area as a tab, in addition to the Execution/Miri/Clippy panels. It seems like an interesting compromise between taking space all the time but being rarely used.

Other possibilities could include:
- having a "custom input" button/checkbox in the top bar (as there is some space available now) which would toggle the input area appearing.
- having some of the panels always visible (but collapsed by default): for this purpose, say, the Execution and Custom Input panels.

letmutx:
Thanks! To answer your questions, I feel

* It should be a text box big enough to show a paragraph.
* The issue had no activity, so I'd assume it's not _that_ desired but could be useful to some people.

I use playground to quickly check my code and run some test cases. But hard coded values in source meant I had to change multiple locations in the source. Now I could automate it with some glue and run my test cases adding more data to stdin.

I'm not sure as to how much waste of space it will be by adding the input button because when more buttons are added they will appear beside the input button, not below or above(I think the picture above is misleading, there won't be any text box on app load: only a button where the `Execution` button is right now). But having a checkbox in the header area to make the input area/button appear is also worth considering. I'd leave it to @shepmaster or someone else to make the decision.

ui/src/sandbox.rs
```diff
@@ -167,9 +176,21 @@ impl Sandbox {
...
-        let output = try!(command.output().map_err(Error::UnableToExecuteCompiler));
+        let mut command = self.execute_command(req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req.edition);
+        let output = if let Some(ref input) = req.stdin {
+            command.stdin(Stdio::piped());
+            command.stdout(Stdio::piped());
+            command.stderr(Stdio::piped());
+            let mut child = try!(command.spawn().map_err(Error::UnableToExecuteCompiler));
+            {
+                let stdin = try!(child.stdin.as_mut().ok_or(Error::UnableToOpenStdin));
+                try!(stdin.write_all(input.as_bytes()).map_err(Error::UnableToWriteInput));
```
Stargateur: Here you could end by having a deadlock if input is > pipe buffer size (~65k in general), so you would wait for write and the child could wait for writing to stdout too and cause no one read the output, both child and parent would be block on writing call. That of course is a very rare scenario in this context.

The solution could be to have a second child that only write the input, so the parent can read the output, or use non blocking pipe, or use select function, etc...

shepmaster:
In case folk hadn't noticed this organically, the playground does now support interactive stdin 😉

PR #434, opened 2018-11-26
Handle when a null value has been serialized

shepmaster:


shepmaster:
The fix has been pushed; because the playground has stronger caching, you will either need to wait 24h or force a reload of the page.

PR #440, opened 2018-12-03
Plumb through errors when loading/saving gists fail

Nemo157:
Just noticed that it looks like it's infinitely loading if you accidentally mess up the gist id, [example link](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=b3354e29612baf458966c4df0916fde3).

One other small change to make the error messages more useful is to fall back to the HTTP status text if a non-JSON error response is returned. Users that actually care about the conversion failure are probably going to be checking the network tab in debug tools to see what's going wrong anyway.

### DIFF
--- a/ui/frontend/Output/Gist.tsx
+++ b/ui/frontend/Output/Gist.tsx
@@ -15,6 +15,7 @@ import {
 import Loader from './Loader';
 
 interface GistProps {
+  error?: string;
   codeUrl?: string;
   gistUrl?: string;
   issueUrl?: string;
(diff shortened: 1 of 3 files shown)


Arnavion:
May also be worth changing the `/meta/gists` endpoint code to not return HTTP 500 in this case. Currently it just returns the error up to `iron` directly. It could inspect the `hubcaps` error for `Fault { NOT_FOUND }` (I assume) and return HTTP 404 instead.

PR #453, opened 2019-01-09
Show autocomplete crates only when the line starts with use

### LINKED ISSUE #452: Only show the crate suggestions when it's the first thing on the line
This should not trigger the crates:

```rust
// ignore this use (oops)
```

CYBAI:
Fix #452

### DIFF
--- a/ui/frontend/AdvancedEditor.tsx
+++ b/ui/frontend/AdvancedEditor.tsx
@@ -10,8 +10,8 @@ const displayExternCrateAutocomplete = (editor: any, autocompleteOnUse: boolean)
   const line = session.getLine(pos.row);
   const precedingText = line.slice(0, pos.column);
 
-  return !!precedingText.match(/extern\s+crate\s*\w*$/) ||
-    (autocompleteOnUse && !!precedingText.match(/use\s+(?!crate|self|super)\w*$/));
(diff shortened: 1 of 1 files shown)


ui/frontend/AdvancedEditor.tsx
```diff
@@ -11,7 +11,7 @@ const displayExternCrateAutocomplete = (editor: any, autocompleteOnUse: boolean)
   const precedingText = line.slice(0, pos.column);
 
   return !!precedingText.match(/extern\s+crate\s*\w*$/) ||
-    (autocompleteOnUse && !!precedingText.match(/use\s+(?!crate|self|super)\w*$/));
+    (autocompleteOnUse && !!precedingText.match(/^use\s+(?!crate|self|super)\w*$/));
```
shepmaster: Would you mind making the same change for `extern crate` (on the line above)?

shepmaster:
Wow, thanks for the super fast PR!

ui/frontend/AdvancedEditor.tsx
shepmaster: And actually, I should have specified that I actually meant "start of line (with optional whitespace)". For example, this should still bring it up:

```rust
fn foo() {
    use 
}
```

shepmaster:
Thank you again!

Issue #429, opened 2018-11-19
miri does not use the correct edition


RalfJung:
In [this example](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=5634a19882ab427ff02dac3fcb675431), "Build" works fine because we selected the 2018 edition, but running it in miri seems to use the 2015 edition.

Is there something miri should change here, or does the playground have to inform miri somehow?

RalfJung:
clippy has the same problem, it also has the borrow checker complain in the above example.

rustfmt does not care, likely because it does not do borrow checking.

shepmaster:
As far as you know, setting `edition = ...` in Cargo.toml is all that Miri needs?

RalfJung:
I have no idea. `cargo miri` is just a small shim that calls `miri`, and it seems possible that it has to add the `--edition` flag itself.

shepmaster:
The most likely place to start poking on the playground side is to get this bit of code (or similar) to trigger:

https://github.com/integer32llc/rust-playground/blob/a7148c804cb2d94f062b44b4bc3e516bd6922a7c/ui/src/sandbox.rs#L473-L475

This touches one of the objectively *worst* pieces of the playground because setting up the Docker environment is a big ol' ball of wax.

shepmaster:
I think that the playground will have some work to do as well in a very similar location. For example, you can't run Clippy on input without a `main`, presumably because I've never passed that information down to the right level.

RalfJung:
A local test indicates that `cargo miri` is doing this correctly.

RalfJung:
Uh actually I may have spoken too fast. I see no output at all any more when I enable the new edition. Investigating.^^

RalfJung:
Argh I just forgot to get libstd with all the MIR.^^ Yes, editions are properly supported by cargo miri, so I think playground is the only missing piece.

RalfJung:
Now that the 2018 edition is the default on the playground, this is even more noticeable.

PR #454, opened 2019-01-10
Make Miri honor selected edition

RalfJung:
Note that I have no idea what I am doing here.  I guessed my way through the Rust part of this until it looked toughly like what compile requests do, and then it got even worse for the TypeScript part as I have never read or written a single line of TypeScript before.^^  I hope this makes some sense... I wouldn't know how to test any of this.

Fixes https://github.com/integer32llc/rust-playground/issues/429


shepmaster:
Thanks!

PR #455, opened 2019-01-12
Update crates available in the playground

dtolnay:
This pulls in a version of Serde that re-exports serde_derive macros, allowing us to upgrade our examples to idiomatic 2018 edition.


shepmaster:
Thanks!

PR #461, opened 2019-01-26
Support edition and crate type when running Clippy

### LINKED ISSUE #382: Running clippy should not require a main function
I tried running clippy on
```rust
use std::fmt::Debug;

pub fn test<T: Debug>(x: Vec<T>) {
    for t in x.iter() {
        println!("{:?}", t);
    }
}
```
Even though "Build" works just fine, clippy complains that there is no `main`.

### LINKED ISSUE #456: Linting code with Clippy does not properly account for the selected Rust edition
# What I did

Ran Clippy on the following Rust 2018 code:

```rust
use base64::STANDARD; // 0.10.0

fn main() {
    println!("{:?}", STANDARD);
}
```

([Playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=3d9ce7f5687f2e278e66c6cfe837b5f3))

# What happened

Clippy complains about a missing external crate:

```
error[E0432]: unresolved import `base64`
 --> src/main.rs:1:5
  |
1 | use base64::STANDARD; // 0.10.0
  |     ^^^^^^ maybe a missing `extern crate base64;`?
``` 

# What I expected to happen

I expected the code to be validly parsed and Clippy warnings to be printed (if any).

# Additional information

The  code compiles/runs fine when using "Run".

shepmaster:
Fixes #456; Fixes #382

### DIFF
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -120,7 +120,7 @@ impl Sandbox {
     pub fn compile(&self, req: &CompileRequest) -> Result<CompileResponse> {
         try!(self.write_source_code(&req.code));
 
-        let mut command = self.compile_command(req.target, req.channel, req.mode, req.crate_type, req.tests, req.backtrace, req);
+        let mut command = self.compile_command(req.target, req.channel, req.mode, req.tests, req.backtrace, req);
 
         let output = try!(command.output().map_err(Error::UnableToExecuteCompiler));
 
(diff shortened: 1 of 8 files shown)


ui/src/sandbox.rs
```diff
@@ -354,13 +357,9 @@ impl Sandbox {
         cmd
     }
 
-    fn miri_command(&self, edition: Option<Edition>) -> Command {
+    fn miri_command(&self, req: impl EditionRequest) -> Command {
         let mut cmd = self.docker_command(None);
-        if let Some(edition) = edition {
-            // set the env var that will cause modify-cargo-toml inside the docker
-            // container to set this in the `Cargo.toml`.
```
RalfJung: I think having some comment about how these pieces work together somewhere is really helpful, it took me a bit to figure this out when I wrote the patch for Miri.^^  The propagation of information through 3-5 layers here is far from obvious.

Issue #457, opened 2019-01-23
Decrease time spent waiting for compilation requests


ffissore:
Have you considered asking the community support in terms of mirroring the playground?
At the moment https://play.rust-lang.org/ takes about ~1.5 secs to compile and run a snippet. Could sharing the load across different VMs make it snappier?

shepmaster:
As of right now, the load average is `0.99, 0.86, 0.86`, so the machine is pretty much right at where it should be. I don't believe I have any longer-term stats handy.

> community support

The EC2 instance provided by Mozilla is 2x2.30GHz with 8GB of RAM, so it's not a tiny machine.

> Could sharing the load across different VMs make it snappier?

It's possible, but would require deeper digging to determine what the problem really is.

For example, the alternate deployment (https://play.integer32.com/) has a load average of `0.05, 0.09, 0.04`, but is a 1x2.40GHz with 1GB of RAM, and compiling takes the same amount of time (1.14s vs 1.19s, measured via Chrome's network inspector).

This *may* mean that the performance bottleneck is mostly centered around disk access.

ffissore:
I've just run the playground on my machine (kind of beefy i7 with 32 gb of ram) and the hello world snippet takes ~1.3 secs, so I've reproduced the issue locally. I'll try to investigate

ffissore:
It looks like the root cause is docker. Starting a temporary container takes ~1sec, regardless of the container and the command you run (try `time docker run --rm ubuntu:18.04 whoami`). Then there's the time taken by cargo.

Only idea that comes to my mind is this.
We start only one container at playground start up time, and have a server running inside the container. Such server exposes an `/execute` API that:
1) makes a copy of the current playground folder into a temporary folder
2) stores the snippet into `src/main.rs`
3) calls `modify-cargo-toml` on the temporary cargo folder
4) runs cargo build/run/whatever
5) deletes the temporary folder

Drawbacks are:
1) an "evil" snippet could list the currently running snippets
2) in order to call the server running inside docker, we need to give network capabilities to it and therefore to the snippets

I'm not linux security expert but maybe the `cap` params currently provided to docker have to be relaxed as well in order to run such server, and this could then lead to more issues

ffissore:
One possible solution to drawback 2 is to run `cargo` through `firejail` (which I've just discovered, so I'm no expert), which has a `--net=none` param

ffissore:
I've tried and failed to implement the solution above.
[...]

PR #463, opened 2019-02-15
Speeding up build/run requests with a pool of idle docker containers

ffissore:
This is a possible solution to #457 

I've sped up the compilation/execution of snippets by using previously started and idle docker containers, rather than starting a new one for each request.
On my dev box, this new version runs a "hello world" in ~600 ms, while the master version takes ~1.3 secs. Since `rustc` takes ~400 ms, the overhead of docker is reduced to ~200 ms per request.

**How it works**
A pool of idle docker containers is started at playground start up. The default size is 10 containers per channel.
By sleeping for a long time in `entrypoint.sh`, containers stay idle.
When a request is handled, a container is removed from the pool. Cargo is called through `cargo.sh`, which cares of exporting env vars and killing the container once it's done. Thus, there's no need to drop a container that has been used.
However idle containers in the pool need to me manually terminated: DockerContainers `Drop` impl takes care of that, and kills them in parallel.
To ensure that `drop` is called when server shuts down, I've used crate `ctrlc` to install a shutdown hook

**Known issues**
Benchmarking the `/execute` api with `siege`, the new *pool* version is slightly slower than the master one: the former replies to ~4.5 req/sec, the latter to ~5.5 req/s
Also, because there's no way to gracefully shutdown iron (see https://github.com/hyperium/hyper/issues/338), if the server is terminated while receiving lots of requests, there's a chance that some docker containers will survive the shutdown: they'll have to be manually terminated.

On a side note, I'm new to rust, so any feedback is welcome

### DIFF
--- a/compiler/base/Dockerfile
+++ b/compiler/base/Dockerfile
@@ -41,6 +41,7 @@ RUN cd / && \
 WORKDIR /playground
 
 ADD Cargo.toml /playground/Cargo.toml
+ADD cargo.sh /playground/cargo.sh
 ADD crate-information.json /playground/crate-information.json
 RUN cargo fetch
 
--- a/compiler/base/cargo.sh
+++ b/compiler/base/cargo.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+set -eu
+
+POSITIONAL=()
+
+while [[ $# -gt 0 ]]; do
+        case $1 in
+                --env)
+                        export $2
+                        shift 2
+                        ;;
+                *)
+                        POSITIONAL+=("$1")
+                        shift
+                        ;;
+        esac
+done
+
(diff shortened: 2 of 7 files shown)


ffissore:
Also, this fix is for `Sandbox` `compile` and `execute` methods only: the other methods (`format`, `clippy`...) are unchanged

shepmaster:
Thank you! As a fair warning, it will probably take me a little while to dig in here. If I haven't usefully commented in a few weeks, feel free to ping me again!

ffissore:
As requested, a kind ping to check this PR. I've just rebased it with the latest master

ffissore:
Hello there, another kind ping to check this PR. I've just rebased it with the latest master

PR #473, opened 2019-03-25
install miri via rustup

RalfJung:
Let's hope this works :D


shepmaster:
Can you believe that @oli-obk thought that I would do this myself? Hah!

PR #481, opened 2019-04-20
Add troubleshooting for Docker mount denied on mac

### LINKED ISSUE #480: Docker mount denied on macOS
While trying to run `rust-playground` locally, I can build the compiler docker images successfully and also launch `ui` folder successfully. However, if I tried to `run`, `build` or `rustfmt` (any commands need docker) in my localhost, I will get following error

```
docker: Error response from daemon: Mounts denied: 
The paths /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/output and /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/input.rs
are not shared from OS X and are not known to Docker.
You can configure shared paths from Docker -> Preferences... -> File Sharing.
See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
.
time="2019-04-19T13:36:22+09:00" level=error msg="error waiting for container: context canceled"
```

I tried to follow the instruction `Docker -> Preferences... -> File Sharing` to set the path but I can't add it because the `/private` path already includes the `/var/folders` that `/private` is default paths for macOS.

After doing some search, I found [this SO answer](https://stackoverflow.com/a/45123074/1863619) looks related but still not figure out how to fix it yet.

CYBAI:
I've closed the issue but this is related to fix  #480 🙇

### DIFF
--- a/ui/README.md
+++ b/ui/README.md
@@ -17,3 +17,28 @@ of choice.
 
 [dotenv]: https://crates.io/crates/dotenv
 [gist]: https://developer.github.com/v3/gists/#authentication
+
+### Troubleshooting
+
+#### macOS
+
+After launching `ui`, when you try to do any action (ex. `build`, `rustfmt`, `run` and so on), you get errors from Docker about "Mounts denied":
+
+```
+docker: Error response from daemon: Mounts denied:
+The paths /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/output and /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/input.rs
+are not shared from OS X and are not known to Docker.
+You can configure shared paths from Docker -> Preferences... -> File Sharing.
+See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
+.
+time="2099-12-31T00:00:00+00:00" level=error msg="error waiting for container: context canceled"
+```
+
+To fix this issue, set the `TMPDIR` environment variable to a path that Docker can mount:
+
+```
+mkdir tmp
+TMPDIR=$PWD/tmp cargo run
+```
+
+(Note: This was reported at [#480](https://github.com/integer32llc/rust-playground/issues/480))


README.md
```diff
@@ -111,6 +111,32 @@ cd compiler
...
+docker: Error response from daemon: Mounts denied:
+The paths /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/output and /var/folders/dx/l5pn75zx5v9_cwstvgwc5qyc0000gn/T/playground.6gEHdGUM6XPU/input.rs
+are not shared from OS X and are not known to Docker.
+You can configure shared paths from Docker -> Preferences... -> File Sharing.
+See https://docs.docker.com/docker-for-mac/osxfs/#namespaces for more info.
+.
+time="2099-12-31T00:00:00+00:00" level=error msg="error waiting for container: context canceled"
+```
+
+Then, you can set `TMPDIR` to fix this issue.
```
shepmaster: Presumably you also need to configure Docker to include the path to the new `TMPDIR` in the set of directories it can use for mounting, as alluded to in the error message?

shepmaster:
```suggestion
To fix this issue, set the `TMPDIR` environment variable to a path that Docker can mount:
```

README.md
```diff
@@ -111,6 +111,32 @@ cd compiler
 ./fetch.sh # If you just want the current playground
 ```
 
+### Troubleshooting
+
+#### Mac
+
+After launching `ui` successfully locally, when you try to do any action (ex. `build`, `rustfmt`, `run` and so on), if you get errors about Docker Mount denied like following
+
+```
```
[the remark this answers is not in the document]
CYBAI: Ok!

README.md
CYBAI: Yeah, you're right. If the user set a `TMPDIR` which is not included in the Docker preference, then he/she will need to set it.

I didn't need to set it because I set the `TMPDIR` under my project path which is under `/Users` and `/Users` is also a default setting in Docker for Mac.

shepmaster:
Thanks!

PR #482, opened 2019-04-20
Make rustfmt honor selected edition

### LINKED ISSUE #464: Rustfmt should run under the selected edition mode
I was looking at https://github.com/rust-lang/rust/issues/58491, and tried to format this working code: https://play.rust-lang.org/?version=nightly&edition=2018&gist=7e6d977543fff0d7b0b2da8c864f1c69

RustFmt, however, gave the parsing error that gets in the 2015 edition.  I'm guessing that means the appropriate edition flags aren't being passed for tools?

CYBAI:
Fix #464

### DIFF
--- a/ui/frontend/actions.ts
+++ b/ui/frontend/actions.ts
@@ -2,7 +2,13 @@ import fetch from 'isomorphic-fetch';
 import { ThunkAction as ReduxThunkAction } from 'redux-thunk';
 import url from 'url';
 
-import { clippyRequestSelector, getCrateType, isAutoBuildSelector, runAsTest } from './selectors';
+import {
+  clippyRequestSelector,
+  formatRequestSelector,
+  getCrateType,
+  isAutoBuildSelector,
+  runAsTest,
+} from './selectors';
 import State from './state';
 import {
   AssemblyFlavor,
@@ -420,6 +426,11 @@ export const gotoPosition = (line, column) =>
 const requestFormat = () =>
   createAction(ActionType.RequestFormat);
 
+interface FormatRequestBody {
+  code: string;
+  edition: string;
+}
+
 interface FormatResponseBody {
   code: string;
   stdout: string;
@@ -438,8 +449,7 @@ export function performFormat(): ThunkAction {
   return function(dispatch, getState) {
     dispatch(requestFormat());
 
-    const { code } = getState();
-    const body = { code };
+    const body: FormatRequestBody = formatRequestSelector(getState());
 
     return jsonPost(routes.format, body)
       .then(json => {
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -977,6 +977,8 @@ mod test {
     }
     "#;
 
+    const BACKTRACE_NOTE: &str = "Run with `RUST_BACKTRACE=1` environment variable to display a backtrace";
+
     #[test]
     fn backtrace_disabled() -> Result<()> {
         let req = ExecuteRequest {
@@ -988,7 +990,7 @@ mod test {
         let sb = Sandbox::new()?;
         let resp = sb.execute(&req)?;
 
-        assert!(resp.stderr.contains("Run with `RUST_BACKTRACE=1` for a backtrace"));
+        assert!(resp.stderr.contains(BACKTRACE_NOTE));
         assert!(!resp.stderr.contains("stack backtrace:"));
 
         Ok(())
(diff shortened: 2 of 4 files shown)


CYBAI:
> What is some behavior that changes when we switch editions? Can we add a test next to where we added `None` that fails for edition 2015?

Yes! We should add a test for it! How about using part of the example in #464 as test?

```rust
fn main() {
    use std::num::ParseIntError;
    let result: Result<i32, ParseIntError> =
        try { "1".parse::<i32>()? + "2".parse::<i32>()? + "3".parse::<i32>()? };
    assert_eq!(result, Ok(6));
}
```

Then we should fail to format it in `Edition2015` and succeed in `Edition2018`.

ui/src/sandbox.rs
```diff
@@ -1082,20 +1082,36 @@ mod test {
...
-            code: "fn foo () { method_call(); }".to_string(),
-            edition: None,
+            code: "fn main() { use std::num::ParseIntError; let result: Result<i32, ParseIntError> = try { \"1\".parse::<i32>()? + \"2\".parse::<i32>()? + \"3\".parse::<i32>()? }; assert_eq!(result, Ok(6)); }".to_string(),
+            edition: Some(Edition::Rust2015),
+        };
+
+        let sb = Sandbox::new().expect("Unable to create sandbox");
+        let resp = sb.format(&req);
+
+        assert!(resp.is_err());
```
CYBAI: Hmmm, I'm not sure why I even cannot pass the original test case in my local but I think logically this test should be passed 😓? Wait for CI result 🙏

ui/src/sandbox.rs
```diff
@@ -1075,19 +1082,36 @@ mod test {
...
         let req = FormatRequest {
-            code: "fn foo () { method_call(); }".to_string(),
+            code: "fn main() { use std::num::ParseIntError; let result: Result<i32, ParseIntError> = try { \"1\".parse::<i32>()? + \"2\".parse::<i32>()? + \"3\".parse::<i32>()? }; assert_eq!(result, Ok(6)); }".to_string(),
+            edition: Some(Edition::Rust2015),
+        };
+
+        let sb = Sandbox::new().expect("Unable to create sandbox");
+        let resp = sb.format(&req).expect("Unable to format code");
+
+        assert!(resp.stderr.contains("error: expected identifier, found `\"1\"`"));
```
CYBAI: I found that I cannot pass the original test case but I think this might be passed logically 😓?
Btw, I just found it seems that we didn't run `cargo test` in Travis 👀?

shepmaster: > we didn't run `cargo test` in Travis

Yeah, I've been putting off handling that because it's going to require a reasonable amount of work to get CI to handle it (the docker containers need to be able to be passed between build steps without leaking the passwords for uploading them to docker hub)

shepmaster:
I [force-pushed some commits](https://github.com/integer32llc/rust-playground/compare/2bf0f1f3ca11dbd06478ae7fe08adbc59fb2d49b..1cf77a78eda592dbdc4233e6174b7c559fd46d18) to fix the tests that broke over time and also tweaked some style of the tests you added (mostly by using `r#""` strings and creating a variable). I then [squashed the unit test commit](https://github.com/integer32llc/rust-playground/compare/1cf77a78eda592dbdc4233e6174b7c559fd46d18..cdfbdcb996215074c65221fcf34b1e06a3499cb8) along with the implementation.

ui/frontend/actions.ts
```diff
@@ -2,7 +2,13 @@ import fetch from 'isomorphic-fetch';
 import { ThunkAction as ReduxThunkAction } from 'redux-thunk';
 import url from 'url';
 
-import { clippyRequestSelector, getCrateType, isAutoBuildSelector, runAsTest } from './selectors';
+import {
+  clippyRequestSelector,
+  formatRequestSelector,
```
shepmaster: Thanks for following this new pattern; I know it can be easy to just leave the code as it was! ❤️

CYBAI:
> I [force-pushed some commits](https://github.com/integer32llc/rust-playground/compare/2bf0f1f3ca11dbd06478ae7fe08adbc59fb2d49b..1cf77a78eda592dbdc4233e6174b7c559fd46d18) to fix the tests that broke over time and also tweaked some style of the tests you added (mostly by using `r#""` strings and creating a variable). I then [squashed the unit test commit](https://github.com/integer32llc/rust-playground/compare/1cf77a78eda592dbdc4233e6174b7c559fd46d18..cdfbdcb996215074c65221fcf34b1e06a3499cb8) along with the implementation.

Thanks for fixing the test! and your patch looks much better 😆

PR #484, opened 2019-04-30
Use && operator for conditional rendering

joekyo:
Two `null`s are removed. :)


shepmaster:
Ooof, this was some early React code, before I even realized you could use `&&` for this!

Looks good, thanks!

PR #512, opened 2019-07-09
Fix highlight for warnings with error number

### LINKED ISSUE #505: Expand warning (and error?) regex to include error codes
```rust
enum MyEnum {
    one,
}

fn main() {
    match MyEnum::one {
        one => {}
        _ => {}
    }
}
```

This warning is not highlighted:

```
warning[E0170]: pattern binding `one` is named the same as one of the variants of the type `MyEnum`
 --> src/main.rs:7:9
  |
7 |         one => {},
  |         ^^^ help: to match on the variant, qualify the path: `MyEnum::one`
```

CYBAI:
Fix #505

| Before | After |
| ------- | ----- |
| image | image |

### DIFF
--- a/ui/frontend/highlighting.ts
+++ b/ui/frontend/highlighting.ts
@@ -7,7 +7,13 @@ export function configureRustErrors({
   reExecuteWithBacktrace,
 }) {
   Prism.languages.rust_errors = { // eslint-disable-line camelcase
-    'warning': /^warning:.*$/m,
+    'warning': {
+      pattern: /^warning(\[E\d+\])?:.*$/m,
+      inside: {
+        'error-explanation': /\[E\d+\]/,
+        'see-issue': /see issue #\d+/,
+      },
+    },
     'error': {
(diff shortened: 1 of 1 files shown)


ui/frontend/highlighting.ts
```diff
@@ -7,7 +7,7 @@ export function configureRustErrors({
   reExecuteWithBacktrace,
 }) {
   Prism.languages.rust_errors = { // eslint-disable-line camelcase
-    'warning': /^warning:.*$/m,
+    'warning': /^warning(\[E\d+\])?:.*$/m,
```
shepmaster: This is a great start!

Do you see a new lines down where it says `'error-explanation': /\[E\d+\]/,`? Can we follow the same pattern here to make the `[E1234]` part a link to the error guide?

CYBAI: Cool! Let me try do update it! Thanks!

shepmaster:
Thanks!

PR #518, opened 2019-07-24
miri no longer needs -Zmiri-seed

RalfJung:
With https://github.com/rust-lang/miri/pull/851, this is the default.


RalfJung:
As always, thanks for hosting Miri. :)  I think this is making a *big* difference in terms of people being able to test snippets of code for UB, and I am so thankful that you are giving us the chance to do this on the playground.

I even mentioned this in a scientific paper I recently submitted. ;) Helps a lot to establish credibility for the interpreter if we can claim that the "official" Rust playground has it integrated. :D

PR #523, opened 2019-07-31
Add config option to disable automatic pairing of characters

### LINKED ISSUE #516: Editor inserts unwanted brackets when typing
Typing `(` inserts `)` after the caret. Typing `[` inserts `]` after the caret. Typing `{` followed by a line break inserts a line with `}` after the caret. I'd say this is a bug, not a feature. It annoys me all the time, as I have to go and clean up unwanted brackets. The end parentheses `)` sometimes takes care of themselves as typing an end parentheses sometimes overwrites the unsolicited one, but sometimes not, such as when typing a start parenthesis before something that has already been written. I guess I'm expected to select the entire expression I want to surround with parentheses, so that the unsolicited end parenthesis goes at the end of the selection, which is of course not what I do. Knowing this workaround also doesn't help when I only want to insert a lone start parenthesis, which can happen when I want to move a parenthesis from one place to another. The end curly brace `}` doesn't take care of itself at all, requiring erasing entire lines after every block I write. The unwanted end curly braces can also appear outside the viewport when typing at the bottom of the viewport, so I don't see them before I compile and get a surprising syntax error.

Nemo157:
Fixes #516

### DIFF
--- a/ui/frontend/AdvancedEditor.tsx
+++ b/ui/frontend/AdvancedEditor.tsx
@@ -2,7 +2,7 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 import { connect } from 'react-redux';
 
 import State from './state';
-import { CommonEditorProps, Crate, Edition, Focus } from './types';
+import { CommonEditorProps, Crate, Edition, Focus, PairCharacters } from './types';
 
 type Ace = typeof import('ace-builds');
 type AceEditor = import('ace-builds').Ace.Editor;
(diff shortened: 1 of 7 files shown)


portstrom:
I was hoping the buggy behavior would be disabled by default, but anyway, this option to make text editing great again is awesome.

shepmaster:
> I was hoping the buggy behavior would be disabled by default,

If you believe there are actual **bugs** with the behavior, it would be appropriate to file an issue with steps to reproduce the bugs with the Ace editor.

Issue #533, opened 2019-09-16
nightly rustc has not been updated for a few weeks


RalfJung:
The "nightly" on playground is from 2019-08-23, more than 3 weeks old. Seems like the nightly update stopped working some time ago?

Cc @shepmaster

RalfJung:
Looking at https://travis-ci.com/integer32llc/rust-playground/jobs/235790592, the issue seems to be that getrandom 0.1.10 got yanked.

PR #534, opened 2019-09-16
getrandom 0.1.10 got yanked, update to 0.1.12

RalfJung:
Hopefully fixes https://github.com/integer32llc/rust-playground/issues/533


shepmaster:
Thanks, but it's not worth spending the 6 hrs of rebuilding the containers for a single crate. I was waiting for getrandom to stabilize before doing this (it had a few yanks). I'll update all the crates.

PR #547, opened 2019-11-20
Add structure crate

srinivasreddy:
This crate helps me in packing/unpacking of integers.


shepmaster:
Please review [our policy][policy] about which crates are accepted. TL;DR: we don't currently accept one-off additions of crates.

[policy]: https://github.com/integer32llc/rust-playground/issues/101

PR #562, opened 2020-01-02
Allow selecting the menu button's text

shepmaster:
You still have to start the drag from off of the button, but it's
better.

### DIFF
--- a/ui/frontend/index.scss
+++ b/ui/frontend/index.scss
@@ -644,6 +644,7 @@ $header-transition: 0.2s ease-in-out;
 
   width: 100%;
   transition: color $header-transition;
+  user-select: text;
 }
 
 .button-menu-item {


shepmaster:
/cc @RalfJung 

selecty

RalfJung:
Cool :)
In Firefox you can alt-click to start selecting text inside a link, but at least without that PR even that does not work the playground.

shepmaster:
> to start selecting text inside a link

These are `<button>` elements, and if you start and end the click inside them, it counts as clicking the button, which closes the menu. If you are on Linux, it might trigger the alternate/middle mouse button clipboard, but on e.g. stock macOS you have to start or end the drag off the button. 😉

RalfJung:
Confirmed that this also works in Firefox. :D

Issue #563, opened 2020-01-08
Async main fn is not detected


nate-onesignal:
When using tokio, it is common to write an `async main` fn wrapped with the `#[tokio::main]` attribute. The playground fails to recognize the generated main fn and issues this warning:

```
No main function was detected, so your code was compiled
but not run. If you’d like to execute your code, please
add a main function.
```

Here is a basic example that fails to run:

```rust
#[tokio::main]
async fn main() {
    println!("Hello");
}
```

shepmaster:
Thanks for filing this. I noticed it earlier but neglected to log it. 

To address this, [the regex detecting the `main` function needs to be updated](https://github.com/integer32llc/rust-playground/blob/5438d93f682f746e00c16fb4953c6c69d3b6c488/ui/frontend/selectors/index.ts#L13-L14).

shepmaster:
Closed via #564

PR #564, opened 2020-01-16
Add support for async main function

drewkett:
This adds support for an `async main` function. It updates the regex and tests per #563. It does not check for `#[tokio::main]` or another macro to see if its a properly configured `main` function.

I have not tried running the tests but it seems like they should pass.


shepmaster:
Thanks!

PR #568, opened 2020-01-22
Update version of yanked time crate

rchaser53:
I'm planning to send the PR about rustfmt.toml, but CI is failed now.
So I send this PR.

### DIFF
--- a/compiler/base/Cargo.toml
+++ b/compiler/base/Cargo.toml
@@ -91,7 +91,7 @@ version = "=0.1.5"
 
 [dependencies.bstr]
 package = "bstr"
-version = "=0.2.9"
+version = "=0.2.10"
 
 [dependencies.build]
 package = "winapi-build"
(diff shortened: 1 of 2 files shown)


shepmaster:
I wouldn't worry about this. Any changes you do are unlikely to involve these containers, and changes to the containers aren't tested as part of pull requests in the first place.

Beyond that, this doesn't update all the references to `time`, and we usually upgrade all the crates at once, since we have to rebuild it from scratch if even one crate changes.

rchaser53:
> I wouldn't worry about this. Any changes you do are unlikely to involve these containers, and changes to the containers aren't tested as part of pull requests in the first place.

Don't I need to send this PR perhaps? If it is so, I will close this PR.

> Beyond that, this doesn't update all the references to time, and we usually upgrade all the crates at once, since we have to rebuild it from scratch if even one crate changes.

Oh sorry. I found the tool named top-crate in this repository. It seems to be used to update the dependencies, so I used the top-crate and updated Cargo.toml and crate-information.json. And then I tried to build the image using build.sh and passed the test.

rchaser53:
chrono 0.4.10 cannot be built on stable, so I didn't change it.
ref: https://github.com/chronotope/chrono/issues/373

shepmaster:
> chrono 0.4.10 cannot be built on stable, so I didn't change it.

Yep, I've been manually holding that back, in the hopes that they fix the issue.

PR #577, opened 2020-02-20
Add ndarry-rand package

liufuyang:
In order to test ndarry conveniently, can we add this `ndarry-rand` lib in the play ground as well? It would be nice to be able to create some random arrays easily :)

For example, I would like to test and run example like this in the playground:
```rust
use ndarray::prelude::*;
use ndarray::Array;
use ndarray_rand::RandomExt;
use ndarray_rand::rand_distr::Uniform;

fn main() {
    let a = Array::random((2, 5), Uniform::new(0., 10.));
    println!("{:8.4}", a);
}
```

### DIFF
--- a/compiler/base/Cargo.toml
+++ b/compiler/base/Cargo.toml
@@ -503,6 +503,10 @@ version = "=0.2.3"
 package = "ndarray"
 version = "=0.13.0"
 
+[dependencies.ndarray_rand]
+package = "ndarray-rand"
+version = "=0.11.0"
+
 [dependencies.net2]
 package = "net2"
 version = "=0.2.33"
--- a/compiler/base/crate-information.json
+++ b/compiler/base/crate-information.json
@@ -605,6 +605,11 @@
     "id": "ndarray"
   },
   {
+    "name": "ndarray-rand",
+    "version": "0.11.0",
+    "id": "ndarray_rand"
+  },
+  {
     "name": "net2",
     "version": "0.2.33",
     "id": "net2"


shepmaster:
https://github.com/integer32llc/rust-playground/blob/master/.github/ISSUE_TEMPLATE.md

liufuyang:
@shepmaster I see. But please allow me to explain why I would like to have it here, as I am trying to create a quick start doc for `ndarry` and it's not merged yet but you can see it here https://github.com/liufuyang/ndarray/blob/add-step-by-step-readme/README-quick-start.md

And in order to make it very close looking to the numpy package guide, for example in this step https://numpy.org/devdocs/user/quickstart.html#shape-manipulation
This `ndarray-rand` is needed. 

I know `ndarray-rand` is not some top list crate for sure but sadly it is not together with `ndarray` otherwise we don't need to add it. But the functionality it provides is pretty tight with `ndarray`. You already support `ndarray`, in that sense I think it would be nice to have it in the play ground.

It's just so easy to try out numpy for example in google co-labs. But it would it make it difficult for people to play around with `ndarry` with out this `ndarray-rand` package.

Also I believe among the most popular 100 crates, many probably is not often used in the playground environment, especially for those ones used for doing web applications or multi-thread computations. But I guess many would use this playground to tryout `ndarray` if they have slight interest to use it.

Don't you agree?

liufuyang:
BTW, as you probably already know, https://github.com/rust-ndarray/ndarray/tree/master/ndarray-rand is in the `ndarray` repo. 

Perhaps that "100 crates rule" should be updated to be a 100 family rule? I a highly used created is closely coupled with some other mall crates, perhaps those small coupled crates should be included as well? Otherwise it is difficult to play in the playground.  It feels like some nice toys are there but their important accessories are missing. @shepmaster

PR #590, opened 2020-04-04
Make import suggestions clickable

### LINKED ISSUE #404: Consider automatically adding suggested includes
```rust
fn example(_: RefCell) {}
```

```
error[E0412]: cannot find type `RefCell` in this scope
 --> src/lib.rs:1:15
  |
1 | fn example(_: RefCell) {}
  |               ^^^^^^^ not found in this scope
help: possible candidate is found in another module, you can import it into scope
  |
1 | use std::cell::RefCell;
  |
```

Clicking on `use std::cell::RefCell;` could perhaps add the `use` statement to the code.

Riateche:
In this PR, the frontend detects import suggestions in compiler output and makes them clickable. On click, the suggestion is inserted at the top of the editor.

playground1
playground2

Inserting imports at the beginning is not ideal. Ideally, the new import should go at the end of the imports block, and the imports block should be separated from items with an empty line. Also it won't work if the compile error originates from a nested module. However, the frontend doesn't seem to have tools for more sophisticated behavior, and this implementation greatly simplifies preparing a working snippet the most common case.

Closes #404.

### DIFF
--- a/ui/frontend/index.tsx
+++ b/ui/frontend/index.tsx
@@ -14,6 +14,7 @@ import {
   editCode,
   enableFeatureGate,
   gotoPosition,
+  addImport,
   performCratesLoad,
   performVersionsLoad,
   reExecuteWithBacktrace,
@@ -48,6 +49,7 @@ const store = createStore(playgroundApp, initialState, enhancers);
 configureRustErrors({
   enableFeatureGate: featureGate => store.dispatch(enableFeatureGate(featureGate)),
   gotoPosition: (line, col) => store.dispatch(gotoPosition(line, col)),
+  addImport: (code) => store.dispatch(addImport(code)),
   reExecuteWithBacktrace: () => store.dispatch(reExecuteWithBacktrace()),
   getChannel: () => store.getState().configuration.channel,
 });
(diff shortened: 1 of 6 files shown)


shepmaster:
Seems reasonable on a quick glance. Please undo your changes to the lockfile; that can be addressed separately.

shepmaster:
> deally, the new import should go at the end of the imports block, and the imports block should be separated from items with an empty line

I agree, but rustfmt should take care of that in general. 

> the frontend doesn't seem to have tools for more sophisticated behavior

It’s true, and there’s an issue for somehow magically using Wasm to provide things like that (someday)

Riateche:
> Please undo your changes to the lockfile; that can be addressed separately.

Sure. Done.

ui/frontend/index.tsx
```diff
@@ -14,6 +14,7 @@ import {
   editCode,
   enableFeatureGate,
   gotoPosition,
+  addCodeToBeginning,
```
shepmaster, line 17: WDYT about calling this `addImport` (and similar changes elsewhere) to match the intended usage? That way, if we can ever have something more capable, we know it’s safe to change.

Riateche: Seems reasonable. I've renamed it.

shepmaster:
Thanks!

PR #616, opened 2020-07-05
fix: fit the playground height to the parent view height

86:
The playground height does not fit into its parent height, so the extra scrolling happens.
This PR changes playground height to 100% to match the parent view height.

Before:
rust_playground_before

After:
rust_playground_after


shepmaster:
Thanks! This seems to be a Chrome-only issue, and it definitely hasn't always been this way. Wonder what changed...

PR #620, opened 2020-07-09
fix some a11y issues

lpmi-13:
these accessibility issues were low-hanging fruit, so addressing in one commit:

- add a lang attribute to the html element
- add a meta description (also counts for the SEO lighthouse score)
- add a landmark (main) to the page

### DIFF
--- a/ui/frontend/index.ejs
+++ b/ui/frontend/index.ejs
@@ -1,11 +1,14 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="description" content="A browser interface to the Rust compiler to experiment with the language" />
     <title><%= htmlWebpackPlugin.options.title %></title>
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,700%7CSource+Code+Pro:400,700&amp;subset=latin-ext" rel="stylesheet">
   </head>
-  <body>
-    <div id="playground"></div>
-  </body>
+  <main>
+    <body>
+      <div id="playground"></div>
+    </body>
+  </main>
 </html>


lpmi-13:
current site accessibilty score:
current site a11y issues

accessibility score with this commit:
with commit

lpmi-13:
(NOTE: lower Best Practice score...from 96 to 83...due to console errors from not running a backend)

ui/frontend/index.ejs
```diff
@@ -1,11 +1,12 @@
...
-<html>
+<html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="description" content="UI for the Rust playground" />
     <title><%= htmlWebpackPlugin.options.title %></title>
     <link href="https://fonts.googleapis.com/css?family=Open+Sans:400,600,700%7CSource+Code+Pro:400,700&amp;subset=latin-ext" rel="stylesheet">
   </head>
-  <body>
+  <main>
```
shepmaster, line 9: Body has CSS rules applied to it; so this seems incomplete.

ui/frontend/index.ejs
```diff
@@ -1,11 +1,12 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="UTF-8" />
+    <meta name="description" content="UI for the Rust playground" />
```
shepmaster, line 5: This doesn’t feel like a *useful* description. Is there Some way we could describe the purpose of the playground or the kinds of things it would be used for?

ui/frontend/index.ejs
lpmi-13: ah, fair point! Had added a `<main>` around the `<body>` tags intially, but thought I could get away without them...will add back in.

ui/frontend/index.ejs
lpmi-13: yeah, I just grabbed this from the `package.json`...how about something like:
"A full-featured Rust runtime in the browser to experiment with any part of the language that you want"

...but I'm totally happy to just put in whatever you think makes the most sense.

shepmaster:
Thanks!

Issue #323, opened 2018-05-08
Add the ability to reset (clear) the current buffer.


vext01:
Hi,

The rust playground remembers the last buffer you were working on. That's very useful, but sometimes I want to start afresh. I can't see a way to do this currently. Could there be a "clear" button which resets the buffer back to the default "hello world"?

Thanks

shepmaster:
I think this is something that *can* be done. A super hacky way of doing this now would be to clear your browser state for the playground, but that clears *everything*.

shepmaster:
If someone starts working on this, a semi-related request I've heard would be to have some "templates" of interesting pieces of code (e.g. a WASM hello world, or maybe a Futures example) to showcase pieces of the ecosystem. That's a bigger step, but might be worth keeping in the back of your mind.

vext01:
> to have some "templates" of interesting pieces of code

sounds good to me.

You could even go as far as a vim-tutor style experience for learning Rust interactively.

But perhaps you should raise a new issue for that.

Thanks

shepmaster:
An extremely non-obvious thing you can do is to erase all the text and reload the page. This will reload the "hello world" example.

hdevalence:
I just tested the suggestion above, and it doesn't seem to work any more -- reloading the page just reloads the empty text now.  Having an explicit "reset" button would be great!

shepmaster:
One potential enhancement: start tracking the last time someone visited. If they haven't visited in "long enough", then proactively offer to reset back to defaults.

PaulRBerg:
I'm posting the default "Hello World" code here to make it easier for posterity to copy and paste it:

```rust
fn main() {
    println!("Hello, world!");
}
```

DavidAntliff:
I _attempted_ to use the Playground via Chrome app on iOS. I basically failed for this reason - my browser had a large example and there’s no way to select all text for mass delete. Deleting character by character is very slow (and reverts to half speed if holding delete down, for unknown reasons). In the end I couldn’t find a way to reset the playground and had to give up.

Incidentally the “hold space to move cursor” technique doesn’t work either - something about the editor window is unfriendly to iOS.

shepmaster:
For mobile users, I encourage using the simple editor:

<img width="345" alt="image" src="https://user-images.githubusercontent.com/174509/231024542-d94141c5-0cea-4566-85ae-5b93a4493639.png">

Neither Ace nor Monaco work well in most mobile environments. The simple editor is just a browser-native text area.

PR #629, opened 2020-09-01
Add a button in the toolbar to reset the editor

corebreaker:
Fixes the issue #323 .

It add a button to reset the content of the editor. The session storage is also updated.

image

### DIFF
--- a/ui/frontend/AdvancedResetMenu.tsx
+++ b/ui/frontend/AdvancedResetMenu.tsx
@@ -1,0 +1,37 @@
+import React, { useCallback } from 'react';
+import { useDispatch, useSelector } from 'react-redux';
+
+import ButtonMenuItem from './ButtonMenuItem';
+import MenuGroup from './MenuGroup';
+
+import * as selectors from './selectors';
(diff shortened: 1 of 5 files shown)


shepmaster:
Oh wow, I never responded to this, and for that I apologize!

I don't believe that this action is common enough to deserve space on the toolbar, which I attempt to keep minimal to avoid overwhelming people.

My vision for this functionality is two-fold:

1. Place a "reset text" (and potentially "reset all settings") button in the configuration menu.
2. Track when the user last visited and if it's long enough ago (exact amount TBD), pop up an offering to reset the text / settings.

CleanCut:
Huh. I thought it was really weird that there wasn't a reset button right about there. I thought it was a UI bug or something, actually. I would expect there to be some easily-findable way to reset the state.

RandyMcMillan:
please add this

RandyMcMillan:
a simple solution is to add a link to an empty public gist

image

shepmaster:
I decided to add a button in the config menu for now. 

image

PR #635, opened 2020-09-21
explicitly use 'cargo miri run'

RalfJung:
The old `cargo miri` does not work any more, the subcommand must be given explicitly.

(The only reason Miri works on the the playground website right now is that it is using an outdated Miri.)


RalfJung:
> (The only reason Miri works on the the playground website right now is that it is using an outdated Miri.)

(And that is likely going to be resolved by https://github.com/rust-lang/miri/pull/1557.)

shepmaster (approved):

PR #643, opened 2020-11-03
Better protect against long-running container execution

shepmaster:


ui/src/sandbox.rs
```diff
@@ -1374,6 +1478,17 @@ mod test {
...
+        #[tokio::main]
+        async fn docker_process_count() -> usize {
+            let mut cmd = docker_command!("ps", "-a");
+            let output = cmd.output().await.expect("Unable to get process count");
+            let output = String::from_utf8_lossy(&output.stdout);
+            // Skip one line of header
+            output.lines().skip(1).count()
+        }
+
+        assert_eq!(0, docker_process_count(), "There must be no running docker processes");
```
pietroalbini, line 1490: I'd prefer if this stores the count and then later ensures it's the same, otherwise it'll be impossible to run the test suite if another, unrelated container is running.

pietroalbini (approved):

ui/src/sandbox.rs
shepmaster: Yes, but no.

The problem is that if another container is running when the test starts (count = 1) then the timeout container starts (count = 2), then the existing container exits (count = 1), then the test would falsely appear to pass (1 == 1) — global resources, right? 

Ensuring it's zero prevents such a case from occurring.

pietroalbini: I'm feeling like this is making the test suite more fragile though. Another possibility would be to have the function return the ID of the container it launched, so the test can ensure no container with that ID exists.

shepmaster: > the test suite more fragile though

Yep, but right now the test suite literally only runs on my laptop and only when I think to run it 😉

If we were to run in CI, I think we’d be able to ensure no other processes are running. 

> return the ID of the container it launched

Yeah, that was my other idea, but it would involve deeper changes to the calling code (to ignore the ID). That’s probably what I’ll do when this becomes a problem.

PR #647, opened 2020-11-14
Stop the MIR source regex from being greedy

camelid:
Otherwise it would highlight too much of the line if there were multiple MIR
sources. For example with [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=30970d3e39b0bfcba0120b04385da47e), it would highlight the line like this:

> let mut _0: \[closure@[src/main.rs:6:5: 6:23 y:i32\]; // return place in scope 0 at src/main.rs:5:19: 5:38]()

### DIFF
--- a/ui/frontend/highlighting.ts
+++ b/ui/frontend/highlighting.ts
@@ -51,7 +51,7 @@ export function configureRustErrors({
   };
 
   Prism.languages.rust_mir = {
-    'mir-source': /src\/.*.rs:\d+:\d+: \d+:\d+/,
+    'mir-source': /src\/[A-Za-z0-9_.\-]+\.rs:\d+:\d+: \d+:\d+/,
   }
 
   Prism.hooks.add('wrap', env => {


camelid:
I was not able to test this locally because I'm getting an error from yarn:

```console
$ yarn
yarn install v1.22.10
[1/5] 🔍  Validating package.json...
error ui@0.0.1: The engine "node" is incompatible with this module. Expected version "^12.13.0". Got "15.1.0"
error Found incompatible module.
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
```

I don't have any experience with yarn, so if you could help me figure out how to fix this, I would really appreciate it!

shepmaster:
> I'm getting an error from yarn

Your node version is too new — the playground is pinned to node 12 (the stable release when I last updated).

shepmaster (approved):
Looks improved, thank you!

image

PR #650, opened 2020-11-17
Update to Node v14.15.1

camelid:
This is the latest LTS version.

### DIFF
--- a/.travis/compile-frontend.sh
+++ b/.travis/compile-frontend.sh
@@ -10,7 +10,7 @@ docker \
     -v $HOME/cache/node/node_modules:/ui/node_modules \
     -v $HOME/cache/node/yarn-cache:/root/.yarn-cache \
     --workdir /ui \
-    node:12.13 \
+    node:14.15 \
     bash -c 'yarn && \
              yarn test && \
              yarn test:lint && \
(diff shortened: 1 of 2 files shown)


camelid:
Even if we don't upgrade all the way to v15, it might make sense to upgrade to v14 (the LTS).

shepmaster:
> it might make sense to upgrade to v14

This is what I'd prefer, yeah. [Even numbered versions](https://nodesource.com/blog/understanding-how-node-js-release-lines-work/) become LTS versions. Since I don't care to know all the intricacies of Node versioning, I've been trying to stay with those. 😉 

> This is the latest version that Homebrew installs for me

Ditto. What I do is:

```
% brew install node@14
```

Then I have a shell alias:

```zsh
alias node14='source ~/bin/node14'
```

That calls the script:

```zsh
PATH="/usr/local/opt/node@14/bin:$PATH"
```

This is pretty janky, and there are real solutions like [nvm](https://github.com/nvm-sh/nvm), but it's just the right amount of simple for me.

camelid:
> Since I don't care to know all the intricacies of Node versioning, I've been trying to stay with those. 😉

That sounds good to me! Yeah, their versioning scheme is not very approachable :)

camelid:
Your solution works like a charm, thank you :)

shepmaster (approved):

shepmaster:
Thanks!

PR #664, opened 2020-12-30
feat: allow importing Bytes with serde enabled

gakonst:
It'd be nice if this were allowed


shepmaster:
You will need to convince the bytes maintainers that this should be enabled, have them add it to their Cargo.toml, then release a new version. See https://github.com/integer32llc/rust-playground/issues/192#issuecomment-388589736 for details.

PR #669, opened 2021-01-09
Detect `fn main<>()`

Havvy:
You can use an empty generics container on any function, including main. Definitely not idiomatic, but also seemingly trivial to support.


shepmaster:
Seems reasonable; could you add a test to ui/frontend/selectors/index.spec.ts?

You happen to be the first PR after I reworked CI a bit, and I see I've broken it for people other than myself; feel free to ignore any test failures except for "Build frontend" and I'll work on fixing those when I can.

fee1-dead:
What about `fn main<>() where {}`? It's cursed but actually valid syntax.

Issue #502, opened 2019-06-13
Allow pretty printing the AST/HIR


Centril:
It would be really neat if there was an option to dump the AST and HIR representations of programs alongside the option to dump the MIR representation. This would be invaluable when working on the compiler.

Centril:
cc https://github.com/integer32llc/rust-playground/issues/371

camelid:
Note: I think the command to use for printing the HIR would be either `-Z unpretty=hir` (a source-like representation) or `-Z unpretty=hir-tree` (more like `Debug` output).

jyn514:
I'm interested in working on this. I'll try to get around to it this weekend.

camelid:
Note that pretty-printing the AST will probably have to wait until `@LeSeulArtichaut` implements `-Z unpretty=ast`.

callanchen:
I am wondering if the project is still being maintained.

PR #683, opened 2021-02-21
Add support for -Zunpretty=hir

jyn514:
This is the first half of https://github.com/integer32llc/rust-playground/issues/502. Adding support for the AST requires first adding `-Zunpretty=ast` to the rust compiler (https://github.com/rust-lang/rust/pull/82304).

I want to run rustfmt on the generated code, but I'm not sure how to do that.. This is what I tried so far:

```diff
diff --git a/ui/src/sandbox.rs b/ui/src/sandbox.rs
index 838d7a0..5fe4887 100644
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -150,7 +150,10 @@ impl Sandbox {
             Some(file) => {
                 if CompileTarget::Hir == req.target {
                     // Run rustfmt on the generated HIR.
-                    format_command(req).arg(file)
+                    let mut rustfmt = self.format_command(req, true);
+                    rustfmt.arg(dbg!(&file));
+                    log::debug!("formatting generated HIR: {:?}", rustfmt);
+                    run_command_with_timeout(rustfmt)?;
                 }
                 read(&file)?.unwrap_or_else(String::new)
             }
@@ -200,7 +203,7 @@ impl Sandbox {
 
     pub fn format(&self, req: &FormatRequest) -> Result<FormatResponse> {
         self.write_source_code(&req.code)?;
-        let command = self.format_command(req);
+        let command = self.format_command(req, false);
 
         let output = run_command_with_timeout(command)?;
 
@@ -356,14 +359,19 @@ impl Sandbox {
         cmd
     }
 
-    fn format_command(&self, req: impl EditionRequest) -> Command {
+    fn format_command(&self, req: impl EditionRequest, single_file: bool) -> Command {
         let crate_type = CrateType::Binary;
 
         let mut cmd = self.docker_command(Some(crate_type));
 
         cmd.apply_edition(req);
 
-        cmd.arg("rustfmt").args(&["cargo", "fmt"]);
+        cmd.arg("rustfmt");
+        if single_file {
+            cmd.arg("rustfmt");
+        } else {
+            cmd.args(&["cargo", "fmt"]);
+        }
 
         log::debug!("Formatting command is {:?}", cmd);
 ```
But it didn't actually format the code, I'm not sure what's going wrong. AFAIK rustfmt formats files in places, not to stdout.

### CODE REVIEW

File: ui/frontend/selectors/index.ts [hunk 1]
```diff
@@ -102,7 +103,7 @@ export const miriVersionDetailsText = createSelector([getMiri], versionDetails);
 
 const editionSelector = (state: State) => state.configuration.edition;
 
-export const isWasmAvailable = (state: State) => (
+export const isNightlyChannel = (state: State) => (
   state.configuration.channel === Channel.Nightly
 );
```
shepmaster, line 108: I'd prefer to have more semantic selectors, so something like:

```suggestion
const isNightlyChannel = (state: State) => (
  state.configuration.channel === Channel.Nightly
);
export const isWasmAvailable = isNightlyChannel;
export const isHirAvailable = isNightlyChannel;
```

shepmaster (commented):
> I want to run rustfmt on the generated code, but I'm not sure how to do that

You won't be able to; not the way you are trying to, at least. Rustfmt is in a completely separate container (there are 6: stable, beta, nightly, miri, clippy, rustfmt). This allows updating the compiler when miri/clippy/rustfmt are broken.

The same general idea is wanted for expanding macros, so any solution may want to attempt to encompass that, but I expect that the easiest way is to make a second request from the frontend.

ui/frontend/selectors/index.ts
jyn514: Sure thing, done.

jyn514:
> You won't be able to; not the way you are trying to, at least. Rustfmt is in a completely separate container (there are 6: stable, beta, nightly, miri, clippy, rustfmt). This allows updating the compiler when miri/clippy/rustfmt are broken.

Hmm, maybe I could do it by feeding the input to rustfmt on stdin instead of trying to use the same files between containers? I don't think the `run_command_with_timeout` API allows for that though, to write to stdin I need to have a `Child` handle: https://doc.rust-lang.org/std/process/struct.Stdio.html#method.piped

jyn514:
FWIW, I think this would be useful to land as is, and I can try to add rustfmt in a follow-up PR.

ui/frontend/BuildMenu.tsx
```diff
@@ -55,7 +57,11 @@ const BuildMenu: React.SFC<BuildMenuProps> = props => {
         Build and show the resulting LLVM IR, LLVM’s intermediate representation.
       </ButtonMenuItem>
       <ButtonMenuItem name="MIR" onClick={compileToMir}>
-        Build and show the resulting MIR, Rust’s intermediate representation.
+        Build and show the resulting MIR, Rust’s intermediate representation for control flow.
```
camelid, line 60:
```suggestion
        Build and show the resulting MIR, Rust’s control-flow-based intermediate representation.
```

ui/frontend/BuildMenu.tsx
```diff
@@ -55,7 +57,11 @@ const BuildMenu: React.SFC<BuildMenuProps> = props => {
...
+      </ButtonMenuItem>
+      <ButtonMenuItem name="HIR" onClick={compileToHir}>
+        Build and show the resulting HIR, Rust’s intermediate representation for items in the current crate.
```
camelid, line 63: "for items in the current crate" is kind of vague. I think something along the lines of

```suggestion
        Build and show the resulting HIR, Rust’s syntax-based intermediate representation.
```

would be clearer.

jyn514: I don't think this is very accurate; or at least, it doesn't make clear the difference between the AST and the HIR. I'm not 100% sure myself what the difference is.

ui/frontend/BuildMenu.tsx
jyn514: I don't think this is a good description because it implies MIR can represent a whole crate. MIR can *only* represent function bodies or const initializers.

ui/frontend/BuildMenu.tsx
camelid: The HIR is a desugared version of the AST IIUC.

ui/frontend/BuildMenu.tsx
camelid: How does my change imply any more than how it was before that MIR can represent a whole crate?

ui/frontend/BuildMenu.tsx
jyn514: If it's desugared, how can it also be syntax-based?

camelid: It's a subset of Rust's syntax. There's no `for` loop, but to my knowledge it's a strict subset of the full Rust syntax.

ui/frontend/BuildMenu.tsx
jyn514: Yeah I don't know what I was thinking lol, this seems fine too.

jyn514:
@shepmaster this is ready for re-review.

shepmaster (approved):

jyn514:
:tada:

PR #700, opened 2021-04-09
Detect `unsafe fn main() { ... }`

r00ster91:
Currently the playground doesn't detect this `main` function:
```rust
unsafe fn main() {}
```


r00ster91:
Never mind, this doesn't actually compile.

shepmaster:
Since the playground is relatively dumb (using regex), you can trick it by adding a newline

```rust
unsafe
fn main() {}
```

You can also force it to run when it's misdetected:

image

PR #704, opened 2021-04-14
Add favicons

imjasonmiller:
From https://github.com/integer32llc/rust-playground/pull/321#issuecomment-656398504:

> My opinion is that it should be distinct but related to the favicon on https://rust-lang.org.

A possible alternative to the Rust language favicon, but using a "P" for playground instead:

Suggested favicon for the Rust Playground

The colors used are those of:
 * Ferris: `#f74c00`
 * Rust Playground background `#e5e5e5`

The vector graphic, while newly created, is based on the original Rust logo, licensed as [`CC-BY`](https://www.rust-lang.org/policies/media-guide).

Generated via https://realfavicongenerator.net, this pull request also includes favicons for:

* **Desktop**
image
* **iOS**
image
* **Android**
image
* **Windows**
image
* **Safari**
image


Canop:
This is admirable.

Using a different color to avoid confusion with the many other sites having a favicon based on the rust cog is an excellent idea.

I can only approve this change which would make ma gain a few dozens minutes every hour when looking for the playground tab among a few dozens millions other tabs.

Stargateur:
Nothing prevent us to change it later, so I think it's a very good start.

ennmichael:
Nice icon

FedericoStra:
Any update on this?

knanao:
I'm looking forward to being merged these changes. 🦀

Rodrigodd:
Also looking forward to this! And if an opinion is welcome, I think it would be better if the icon has the same color as the Run button (`#aa4422`), it would match the site better.

DavidAntliff:
What can we do to get this merged? I use a lot of titleless bookmarks (huge space saving, try it sometime!), thus relying on the favicon to identify them. This site stands out as missing a favicon.

PR #730, opened 2021-07-06
Allow passing an argument to top-crates for the base directory

chazkiker2:
> Note: if the maintainers/owners of `rust-playground` had any interest in publishing `rust-playground/top-crates` to crates.io, that would be super appreciated and this PR would make it really easy to do that!! 

### Description

- [x] Add support for passing in a custom path argument via the command line to allow for re-use of the top-crates code in other places
- [x] If no args are given, keep the default path of `../compiler/base` to maintain any backwards compatibility within `rust-playground`

### Reasoning

Over at [`optopodi`](https://github.com/optopodi/optopodi), we're hoping to utilize the code found in `top-crates/src/main.rs` to generate the `crate-information.json` automatically — this PR would allow for others to easily reuse this code, too!

### DIFF
--- a/top-crates/src/main.rs
+++ b/top-crates/src/main.rs
@@ -19,6 +19,7 @@ use std::{
     collections::{BTreeMap, BTreeSet, HashSet},
     fs::File,
     io::{Read, Write},
+    path::{Path, PathBuf},
 };
 
 const PLAYGROUND_TARGET_PLATFORM: &str = "x86_64-unknown-linux-gnu";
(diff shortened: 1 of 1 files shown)


shepmaster:
Thanks! I squashed your commits into one consistent story and removed the unrelated formatting changes. I then applied a follow-up commit that removes extra allocation and uses the `Path` / `PathBuf` types instead of strings for path manipulation.

This seems to work well here; care to try this version of the code and see if it still works in your use case?

chazkiker2:
@shepmaster it works wonderfully for my use case, thanks so much!

and sorry for the unnecessary formatting changes — I wasn't planning to push up until Niko mentioned that you may be willing to publish to crates.io!! didn't mean to make extra work for you 😉

shepmaster:
It appears that you've... recreated all of the unrelated formatting changes I removed. Did you attempt to make any meaningful changes or should I just remove them?

chazkiker2:
🤦 Oh my goodness, I did not mean to push that up. 

No, I didn't change a thing.

shepmaster:
Thanks!

PR #733, opened 2021-07-14
Split top-crates into a basic library

shepmaster:


shepmaster:
Well, nightly Cargo is broken, so I'm going to merge this anyway and deal with any fallout when it's fixed.

Issue #357, opened 2018-07-13
RLS Integration


skyne98:
Partially realted to #188 

Hey there, folk! Should we try to compile Rust Language Server into WASM to run it in the playground to provide, at least, the intellisence and completions?

I don't really know how closely it is tied to the rust compiler itself, which cannot be, as far as I know, be compiled for the web at the moment. I think we really need someone with a good pack of knowledge here 😄 

Performance might be an issue, but it should be possible to, in theory, run the RLS in a separate WebWorker. Maybe even compile it to use multiple of them using Emscripten and enabling SharedArrayBuffer support?

shepmaster:
> Should we try to compile Rust Language Server into WASM to run it in the playground

I'm about 99.99% certain that this will simply not be feasible. At a bare minimum, the compiled version of the RLS that I have is **20 MiB**. That is a really hefty download to require of anyone, especially those on metered connections.

I also believe that you'd have to get a huge amount of the compiler to target WASM, which includes a large number of dependencies.

I will not be attempting this, but if someone wishes to take on this Herculean task and is able to show a decent proof-of-concept, of course we'd be interested.

skyne98:
@shepmaster, I think that enabling RLS assistance should be an optional feature, which could be enabled by pressing a button "Run RLS" or something which will download it and run in a worker. This way users will have to download it only when they want to. Also, 20 MiB is not too big of a size for a browser-integrated semi-environment, which will have to be downloaded once.

About dependencies -- even though I have no idea about the parser and analyzer child packages, theoretically they shouldn't require any "native" code or syscalls.

shepmaster:
> which will have to be downloaded once

It will have be be downloaded every 6 weeks for stable, every few weeks for beta, every day for nightly. If you use all of those at the first time, you'd get 60 MB.

But sure, if it's feasible, then it can be optional, that's not *much* of a problem. The Ace editor is already an optional download.

skyne98:
@shepmaster, I can totally understand your scepticism, but it would be super nice to have a snippet sharing full fledged editor for rust. It will allow new users to experiment with rust without needing to install it or do it blindly in the editor. Also, it will allow experienced developers to write and share snippets much faster.

DJMcNab:
Related: https://internals.rust-lang.org/t/2019-strategy-for-rustc-and-the-rls/8361/62. This is the discussion about getting https://github.com/rust-analyzer/rust-analyzer working in the playground.

vultix:
[...]

PR #736, opened 2021-08-21
Rename advanced editor to Ace and add Monaco

### LINKED ISSUE #450: Consider using monaco-editor in place of ace
[Monaco](https://microsoft.github.io/monaco-editor/) is an embedded editor that is used by VSCode and [TypeScript playground](https://www.typescriptlang.org/play/).

## Why?

* It has more features out-of-the-box (such as right-click context menu, command palette <kbd>F1</kbd>, etc.).
* It is extensible (you can add themes, IntelliSense and whatnot).
* In theory, Rust Language Server should work well with Monaco as it does in VS Code. (related to https://github.com/integer32llc/rust-playground/issues/357)

HKalbasi:
This is first phase of #357 

It should have zero impact to users (except 2 or 3 kilobytes more download). If people want to use it they should manually opt-in.

Closes #450


HKalbasi:
@shepmaster Do you disagree with this? Should I close this PR?

ui/frontend/types.ts
```diff
@@ -36,7 +36,8 @@ export interface CommonEditorProps {
 
 export enum Editor {
   Simple = 'simple',
-  Advanced = 'advanced',
```
shepmaster, line 39: Doing it this way will require changes to local_storage.ts to preserve people's current selections. If you don't feel like making those changes, you could also leave the value as `advanced` and solely add monaco.

ui/frontend/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+    monaco.languages.setLanguageConfiguration(modeId, config);
+    monaco.languages.setMonarchTokensProvider(modeId, grammar);
+  });
+};
+
+const MonacoEditor: React.SFC<CommonEditorProps> = props => {
+  return (
+    <MonacoReact
+      language={modeId}
+      theme="vscode-dark-plus"
```
shepmaster, line 76: Do we expect that this will be configurable in the future?

ui/frontend/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,86 @@
+import React from 'react';
+import { CommonEditorProps, Position } from '../types';
+import MonacoReact, { Monaco } from "@monaco-editor/react";
+
+import styles from './Editor.module.css';
+import { config, grammar } from './rust_monaco_def';
+
+class CodeByteOffsets {
```
shepmaster, line 8: I haven't actually looked, but is the same as the other `CodeByteOffsets`?

ui/frontend/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+      { token: 'variable', foreground: '9CDCFE' },
+      { token: 'support.function', foreground: 'DCDCAA' },
+    ]
+  });
+  monaco.languages.register({ // language for editor
+    id: modeId,
+  });
+
+  monaco.languages.onLanguage(modeId, async () => {
+    console.log(modeId);
```
shepmaster, line 65: Can be removed.

ui/frontend/editor/MonacoEditor.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+    // Add one to account for the newline we split on and removed
+    return precedingLines.map(l => l.length + 1).reduce((a, b) => a + b);
+  }
+}
+
+const modeId = 'my-rust';
+
+const initMonaco = (monaco: Monaco) => {
+  monaco.editor.defineTheme('vscode-dark-plus', {
```
shepmaster, line 51: Why do we need to create our own theme? What isn't provided by upstream?

ui/frontend/editor/Editor.tsx
```diff
@@ -118,7 +119,7 @@ const Editor: React.SFC = () => {
   const execute = useCallback(() => dispatch(actions.performPrimaryAction()), [dispatch]);
   const onEditCode = useCallback((c) => dispatch(actions.editCode(c)), [dispatch]);
 
-  const SelectedEditor = editor === EditorType.Simple ? SimpleEditor : AdvancedEditor;
+  const SelectedEditor = editor === EditorType.Simple ? SimpleEditor : editor === EditorType.Monaco ? MonacoEditor : AceEditor;
```
shepmaster, line 122: My preference once we get > 2 items is to make a little lookup object with the enum as keys.

ui/frontend/editor/rust_monaco_def.ts
```diff
@@ -0,0 +1,142 @@
+export const config = {
```
shepmaster, line 1: Does this mean that there's no upstream Rust configuration for Monaco?

shepmaster (commented):
Took a quick pass

shepmaster:
> Do you disagree with this? Should I close this PR?

Nope and nope! Life is busy, is all.

ui/frontend/editor/MonacoEditor.tsx
HKalbasi: Removed.

ui/frontend/editor/MonacoEditor.tsx
HKalbasi: VSCode themes are not compatible with monaco because they use textmate convention for tokens but monaco has its own (funnily, monaco is directly built from vscode source code) so upstream won't provide vscode themes. Original themes are ugly (IMO) and people are more familiar with vscode theme.

That theme is not exactly equal to vscode's dark-plus, but it is close enough and I think it is good for default theme (and user can change theme). But we can remove those nine lines if there is something wrong with them.

ui/frontend/editor/rust_monaco_def.ts
HKalbasi: There is, but I'm not happy with it. Some can be solved in upstream (for example not auto closing ' in lifetimes, or more detailed grammar) and some of changes are related to difference of textmate and monarch tokens so won't be accepted in upstream.

ui/frontend/editor/MonacoEditor.tsx
HKalbasi: I made it configurable.

ui/frontend/types.ts
HKalbasi: I added a migration to local_storage.ts.

ui/frontend/editor/MonacoEditor.tsx
HKalbasi: It should be removed.

shepmaster:
This appears to load JS resources from a CDN:

image

Please adjust the imports so that the JS is served by us.

shepmaster:
There appear to be some commands that won't do anything:

image

Please remove these to avoid bugs being filed that they "don't work".

shepmaster:
Setting the editor to `simple` and then reloading the page appears to still download some (relatively hefty!) parts of Monaco:

image

Please ensure that editor-specific code is only fetched when the editor is selected.

HKalbasi:
> There appear to be some commands that won't do anything

It seems that special commands can't be removed. Whole command palette can be removed, but most of command in it works. Even those commands you mentioned work technically, as they work in vscode. Vscode still shows this command when there is no problem and it does nothing when we click (exactly like here). When we add RA, some problems are added so it comes out of the "always does nothing" mode (still if I find a way to remove commands, I'll remove these commands). So I think keeping command palette is better than removing it completely?

CDN and migration should be solved now.

shepmaster:
> I'm still getting bigger bundles. What do you mean by "including twice"?

https://github.com/integer32llc/rust-playground/pull/736/files#diff-605f974ae4f9dbb24828440f6c62968f21ab63a029efcef60f59d7ff142bfba9R4703-R4711

There are two versions of monaco-editor. The versions need to be unified.

HKalbasi:
Great!

> The next step will be adding some automated UI tests.

The ruby ones? I don't know ruby, but I can copy existing ones. Also I'm not sure what need to be tested / can be tested.

HKalbasi:
@shepmaster ping

shepmaster:
@HKalbasi thanks for the ping. I've rebased and added some small tweaks. There's a [unit test](https://github.com/integer32llc/rust-playground/pull/736/files#diff-f72d89eb46c2b337c312dc6cbef8330c87172d9bd04af7790daf3a0064c54aff) for the migration and an [integration test](https://github.com/integer32llc/rust-playground/pull/736/files#diff-7b597f2b53ba12c9e8d4045b9fd64f72a82e84c6624a38295251b96495d02d11) using Monaco.

Care to pull it locally and see if everything continues to work as you expect?

HKalbasi:
Everything looks good

shepmaster:
Let’s go!

lnicola:
For anyone who wants to give it a try, this is already live on https://play.rust-lang.org/.

shepmaster:
Generally things should always be live <1hour after merging. You may need to aggressively reload the page to break the caching, otherwise waiting an hour or day should work.

Zerotask:
Is there a way to set the editor via GET param, e. g. `editor=monaco`?

bjorn3:
I don't think it should be. All configuration options are personal preferences and shouldn't be overridable by playground links if you don't pay attention to the full url.

lnicola:
@HKalbasi do you know if it's possible to implement the `Ctrl-Enter` shortcut for `Run` in Monaco?

PR #739, opened 2021-09-20
♪ EditiooOoOOnnN ! ♫

lqd:
👋 

The 2021 edition will become stable when 1.56 is released, on the [21st of october](https://forge.rust-lang.org/#current-release-versions), and this draft PR prepares the playground to using it as the default edition after its release. 

Specifically, I've:
- removed the `PLAYGROUND_FEATURE_EDITION2021` handling: it currently emits a warning since the cargo feature is now stable on nightly and beta.
- Set 2021 as the default in the UI.
- Removed the aside when setting the 2021 edition, as it will be available everywhere then.

I'm opening this PR as draft:
- I couldn't really test the server side, and would like CI to run tests for me
- I wanted to ask for advice from Jake, in case I've forgotten other important steps and so on.
- this PR should probably only be merged in a month when 1.56 is released

♪ Edition! ♫


shepmaster:
One thing I think we should also add is a [notification popup](https://github.com/integer32llc/rust-playground/blob/master/ui/frontend/Notifications.tsx) to let people know that 2021 is live and the default. 

We should:

1. Remove the old notification
2. Add a notification for 2021

I don't have a great idea for the notification text or silence date, though.

lqd:
Yes, indeed, good point about the notification. You had added one last time, for the 2018 edition, so I've used the same template. That one had a silence date of 01/01/2019.

Thanks for triggering CI as well, the integration tests should start passing on 1.56 IIUC.

shepmaster:
Could you submit `remove special case for cargo "edition2021" feature` as a separate PR? That should be able to be merged before the next Rust release.

lqd:
Sure, done in https://github.com/integer32llc/rust-playground/pull/743.

shepmaster:
FWIW, I locally renamed beta to stable and ran the tests. I fixed up the failures and force pushed this branch.

lqd:
Thanks @shepmaster 🙏

lqd:
🚀

shepmaster:
If GitHub ever lets the images be pushed…

PR #756, opened 2021-12-09
Add a notification for the Rust 2021 survey

shepmaster:


shepmaster:
Merging as it looks like Miri isn’t currently available and the survey is time-limited.

PR #767, opened 2022-01-11
Use Cargo's JSON diagnostics for suggestions

### LINKED ISSUE #688: [Feature]: Auto adding of use imports always goes above inner attributes
If in the playground you use a type that is not in scope, the compiler helpfully outputs some options to use:

Screen Shot 2021-03-09 at 8 36 07 AM

However, if you click on one of those (and you happen to be using some `inner attributes` such as `#[no_std]` or a `#[feature(...)]` then that `use` statement is still added before that attribute and the compiler complains about it the next time you build/run.

It would be great if the `use` statement was added after the attribute.

hafeoz:
Fix #688 .

Uses JSON message format to communicate with `cargo` when building and running. This allows suggestions to be prompted correctly.

#### Currently

a

#### New output

b

### Possible TO-DOs
- [ ] Any mode other than "building" and "running" is still using human-readable mode since I'm not sure how are they different from the normal cargo output.
- [ ] `cargo_metadata` crate has been used. Not clear if the library can work with different cargo versions, or we need to do some kind of versioning. As far as I tested it works for all three variants at least for now.
- [ ] No multi-line diagnostics have been tested. I wasn't able to find or think of any fixes with more than one line.
- [ ] The standard error output's timeline is not correct. Currently, all cargo's output is located at the bottom of the standard error output, regardless of the time arrangement between outputs. [Related cargo comment](https://github.com/rust-lang/cargo/pull/3000#issuecomment-251723533)
- [ ] Better prompt. Maybe discard the "rendered_message" and render the diagnostics message ourselves? This way we can have suggestions in message clickable, instead of having a list of suggestions *under* the message, which might be confusing.

### DIFF
--- a/ui/frontend/reducers/code.ts
+++ b/ui/frontend/reducers/code.ts
@@ -19,8 +19,28 @@ export default function code(state = DEFAULT, action: Action): State {
     case ActionType.AddMainFunction:
       return `${state}\n\n${DEFAULT}`;
 
-    case ActionType.AddImport:
-      return action.code + state;
+    case ActionType.ApplySuggestion:
+      let state_lines = state.split("\n");
+      let startline = action.startline - 1;
+      let endline = action.endline - 1;
+      let startcol = action.startcol - 1;
+      let endcol = action.endcol - 1;
+      if (startline == endline) {
+        state_lines[startline] = state_lines[startline].substring(0, startcol) + state_lines[startline].substring(endcol);
+      } else {
+        if (state_lines.length > startline) {
+          state_lines[startline] = state_lines[startline].substring(0, startcol);
+        }
+        if (state_lines.length > endline) {
+          state_lines[endline] = state_lines[endline].substring(endcol);
+        }
+        if (endline - startline > 1) {
+          state_lines.splice(startline + 1, endline - startline - 1);
+        }
+      }
+      state_lines[startline] = state_lines[startline].substring(0, startcol) + action.suggestion + state_lines[startline].substring(startcol);
+      state = state_lines.join('\n');
+      return state;
 
     case ActionType.EnableFeatureGate:
       return `#![feature(${action.featureGate})]\n${state}`;
--- a/ui/frontend/reducers/code.ts
+++ b/ui/frontend/reducers/code.ts
@@ -20,13 +20,14 @@ export default function code(state = DEFAULT, action: Action): State {
       return `${state}\n\n${DEFAULT}`;
 
     case ActionType.ApplySuggestion:
-      let state_lines = state.split("\n");
-      let startline = action.startline - 1;
-      let endline = action.endline - 1;
-      let startcol = action.startcol - 1;
-      let endcol = action.endcol - 1;
+      const state_lines = state.split('\n');
+      const startline = action.startline - 1;
+      const endline = action.endline - 1;
+      const startcol = action.startcol - 1;
+      const endcol = action.endcol - 1;
       if (startline == endline) {
-        state_lines[startline] = state_lines[startline].substring(0, startcol) + state_lines[startline].substring(endcol);
+        state_lines[startline] = state_lines[startline].substring(0, startcol) +
+          state_lines[startline].substring(endcol);
       } else {
         if (state_lines.length > startline) {
           state_lines[startline] = state_lines[startline].substring(0, startcol);
@@ -38,7 +39,8 @@ export default function code(state = DEFAULT, action: Action): State {
           state_lines.splice(startline + 1, endline - startline - 1);
         }
       }
-      state_lines[startline] = state_lines[startline].substring(0, startcol) + action.suggestion + state_lines[startline].substring(startcol);
+      state_lines[startline] = state_lines[startline].substring(0, startcol) +
+        action.suggestion + state_lines[startline].substring(startcol);
       state = state_lines.join('\n');
       return state;
 
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -602,7 +602,7 @@ fn parse_json_output(output: Vec<u8>) -> Result<(String, String)> {
                 message,
                 ..
             }) => {
-                composed_stderr_string.push_str(&parse_diagnostic(message));
+                composed_stderr_string.push_str(&parse_diagnostic(message, true));
             }
 
             _ => {}
@@ -612,13 +612,18 @@ fn parse_json_output(output: Vec<u8>) -> Result<(String, String)> {
     Ok((composed_stdout_string, composed_stderr_string))
 }
 
-fn parse_diagnostic(diagnostic: cargo_metadata::diagnostic::Diagnostic) -> String {
+fn parse_diagnostic(
+    diagnostic: cargo_metadata::diagnostic::Diagnostic,
+    should_output_message: bool,
+) -> String {
     let mut diagnostic_string = String::new();
 
-    if let Some(rendered_msg) = diagnostic.rendered {
-        diagnostic_string.push_str(&rendered_msg);
-    } else {
-        diagnostic_string.push_str(&diagnostic.message);
+    if should_output_message {
+        if let Some(rendered_msg) = diagnostic.rendered {
+            diagnostic_string.push_str(&rendered_msg);
+        } else {
+            diagnostic_string.push_str(&diagnostic.message);
+        }
     }
 
     for span in diagnostic.spans {
@@ -636,6 +641,10 @@ fn parse_diagnostic(diagnostic: cargo_metadata::diagnostic::Diagnostic) -> Strin
             "\n[[Line {} Col {} - Line {} Col {}: {}]]",
             span.line_start, span.column_start, span.line_end, span.column_end, label
         ));
+    }
+
+    for children in diagnostic.children {
+        diagnostic_string.push_str(&parse_diagnostic(children, false));
     }
 
     diagnostic_string
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -78,6 +78,8 @@ pub enum Error {
     UnableToReadOutput { source: io::Error },
     #[snafu(display("Unable to read crate information: {}", source))]
     UnableToParseCrateInformation { source: ::serde_json::Error },
+    #[snafu(display("Unable to parse cargo output: {}", source))]
+    UnableToParseCargoOutput { source: io::Error },
     #[snafu(display("Output was not valid UTF-8: {}", source))]
     OutputNotUtf8 { source: string::FromUtf8Error },
     #[snafu(display("Output was missing"))]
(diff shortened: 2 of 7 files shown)


hafeoz:
Should we leave the cargo message in `stdout` instead of moving it to `stderr`? This way the timeline will not be disrupted.

shepmaster:
I just wanted to swing by to let you know that I've seen this PR, but I'm not sure when I'll have time to give it the review focus it deserves. I'm super excited about this change, however!

Random thoughts that may or may not be useful or even relevant (as I haven't even looked at the implementation)...

We'll want to (as best we can!) preserve the structure of the API response/requests. e.g. we might want to add a new structured field for the compiler response, instead of repurposing the existing fields. Sadly, people use the existing APIs and assume they will be stable, even though almost no one has ever asked. 😢 

> Maybe discard the "rendered_message" and render the diagnostics message ourselves? This way we can have suggestions in message clickable, instead of having a list of suggestions _under_ the message, which might be confusing.

Yeah, I think this is the ultimate end goal. Obviously it's more of a PITA, but would allow removing the custom regex-based error / warning formatting as well. Maybe this would also allow splitting out Cargo's output from the programs stderr/stdout as well.

> Not clear if [cargo_metadata] can work with different cargo versions

I'd expect that as long as we are using the `v1` of the output (our whatever is the current version) that it should be compatible.

hafeoz:
> We'll want to (as best we can!) preserve the structure of the API response/requests

So should we preserve (at least to our best) the `stdout` and `stderr` field in the JSON response for URL `/execute`? If that's the case, then we might need to add a new field to the response (like `actions`):

<details>
  <summary>Current response example</summary>
  
```json
{
  "success": false,
  "stdout": "",
  "stderr": "   Compiling playground v0.0.1 (/playground)\nerror[E0433]: failed to resolve: use of undeclared type `HashMap`\n --> src/main.rs:3:13\n  |\n3 |     let a = HashMap::new();\n  |             ^^^^^^^ not found in this scope\n  |\nhelp: consider importing one of these items\n  |\n1 | use hashbrown::HashMap;\n  |\n1 | use std::collections::HashMap;\n  |\n\nFor more information about this error, try `rustc --explain E0433`.\nerror: could not compile `playground` due to previous error\n"
}
```
</details>

<details>
  <summary>Proposed response example</summary>
  
```json
{
  "success": false,
  "stdout": "",
  "stderr": "   Compiling playground v0.0.1 (/playground)\nerror[E0433]: failed to resolve: use of undeclared type `HashMap`\n --> src/main.rs:3:13\n  |\n3 |     let a = HashMap::new();\n  |             ^^^^^^^ not found in this scope\n  |\nhelp: consider importing one of these items\n  |\n1 | use hashbrown::HashMap;\n  |\n1 | use std::collections::HashMap;\n  |\n\nFor more information about this error, try `rustc --explain E0433`.\nerror: could not compile `playground` due to previous error\n",
  "actions": [
    {
      "link": {
        "startline": 9,
        "endline": 9,
        "startcol": 4,
        "endcol": 27,
        "target": "stderr"
      },
      "solution": {
        "startline": 1,
        "endline": 1,
        "startcol": 1,
        "endcol": 1,
        "replacement": "use hashbrown::HashMap;"
      }
    },
    {
      "link": {
        "startline": 9,
        "endline": 9,
        "startcol": 4,
        "endcol": 34,
        "target": "stderr"
      },
      "solution": {
        "startline": 1,
        "endline": 1,
        "startcol": 1,
        "endcol": 1,
        "replacement": "use std::collections::HashMap;"
      }
    }
  ]
}
```
</details>

It's definitely going to be more complex than the implementation this PR proposed (especially for front-end). An advantage for such response is that we might be able to remove almost all regex (not only the "import" one) since highlighting and error location (currently handled [here](https://github.com/integer32llc/rust-playground/blob/master/ui/frontend/highlighting.ts#L13-L31)) can be extracted directly from Cargo JSON output and passed to front-end using the new field in HTTP JSON response.

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
...
+        }
+    }
+
+    Ok((composed_stdout_string, composed_stderr_string))
+}
+
+fn parse_diagnostic(
+    diagnostic: cargo_metadata::diagnostic::Diagnostic,
+    should_output_message: bool,
+) -> String {
```
shepmaster, line 618: Instead of building a brand new string here, it'd be a bit more efficient to accept a `&mut String` argument and append to it directly.

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
...
+    let mut diagnostic_string = String::new();
+
+    if should_output_message {
+        if let Some(rendered_msg) = diagnostic.rendered {
+            diagnostic_string.push_str(&rendered_msg);
+        } else {
+            diagnostic_string.push_str(&diagnostic.message);
+        }
```
shepmaster, line 626: Not a required changed, but this could be something along the lines of this (maybe with some `.as_ref` or `as_deref` calls added):

```suggestion
	let x = diagnostic.rendered.or(&diagnostic.message)
        diagnostic_string.push_str(x);
```

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
...
+    }
+
+    for span in diagnostic.spans {
+        if span.file_name != "src/lib.rs" && span.file_name != "src/main.rs" {
```
shepmaster, line 630: My brain is slow at parsing this. Maybe we could use

```suggestion
        if !matches!("src/lib.rs" | "src/main.rs", span.file_name) {
```

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
...
+        let label = if let Some(label) = span.suggested_replacement {
+            label
+        } else {
+            continue;
+        };
+
+        diagnostic_string.push_str(&format!(
+            "\n[[Line {} Col {} - Line {} Col {}: {}]]",
+            span.line_start, span.column_start, span.line_end, span.column_end, label
+        ));
```
shepmaster, line 643: Instead of creating the intermediate `String` via `format!`, you can format directly to the `&mut String` using `write!`. See also [How can I append a formatted string to an existing String?](https://stackoverflow.com/q/28333612/155423)

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
...
+    let mut composed_stdout_string = String::new();
+
+    let metadata_stream = cargo_metadata::Message::parse_stream(&output[..]);
+
+    for msg in metadata_stream {
+        let message = msg.context(UnableToParseCargoOutputSnafu)?;
+
+        match message {
+            cargo_metadata::Message::TextLine(line) => {
+                composed_stdout_string.push_str(&(line + "\n"))
```
shepmaster, line 598: The same idea for using `write!` can be applied here.

ui/src/sandbox.rs
```diff
@@ -571,6 +584,72 @@ fn build_execution_command(
     cmd
 }
 
+fn parse_json_output(output: Vec<u8>) -> Result<(String, String)> {
+    let mut composed_stderr_string = String::new();
+    let mut composed_stdout_string = String::new();
+
+    let metadata_stream = cargo_metadata::Message::parse_stream(&output[..]);
```
shepmaster, line 591: I'd have expected you could elide the `[..]`:

```suggestion
    let metadata_stream = cargo_metadata::Message::parse_stream(&output);
```

ui/src/sandbox.rs
```diff
@@ -78,6 +78,8 @@ pub enum Error {
     UnableToReadOutput { source: io::Error },
     #[snafu(display("Unable to read crate information: {}", source))]
     UnableToParseCrateInformation { source: ::serde_json::Error },
+    #[snafu(display("Unable to parse cargo output: {}", source))]
```
shepmaster, line 81:
```suggestion
    #[snafu(display("Unable to parse Cargo output: {}", source))]
```

ui/frontend/reducers/code.ts
```diff
@@ -19,8 +19,30 @@ export default function code(state = DEFAULT, action: Action): State {
...
+          state_lines[endline] = state_lines[endline].substring(endcol);
+        }
+        if (endline - startline > 1) {
+          state_lines.splice(startline + 1, endline - startline - 1);
+        }
+      }
+      state_lines[startline] = state_lines[startline].substring(0, startcol) +
+        action.suggestion + state_lines[startline].substring(startcol);
+      state = state_lines.join('\n');
+      return state;
```
shepmaster, line 45: This logic is tricky enough that I can't just glance at it and feel very confident about what it's doing. It'd be helpful to extract out the core logic here and write some unit tests to exercise the various `if`s. Something like this sketch:

```ts
    case ActionType.ApplySuggestion:
       const startline = action.startline - 1;
       const endline = action.endline - 1;
       const startcol = action.startcol - 1;
       const endcol = action.endcol - 1;
       return someLogic(state);
```

shepmaster (commented):
This is an encouraging start, thanks! I added comments, but many of them are around the piece of code that's most likely to change as we push more structured information to the frontend, so don't feel like you need to adhere to it now.

hafeoz:
I've been poking around and writing custom diagnostic renderer for the last few days and it's way more complicated than I expected. In order to not break existing API, we basically need to port [this](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/emitter/struct.EmitterWriter.html) so that it can not only output formatted message but also "actions" linked to part of the message. A port would be at least 1000+ lines and include some dirty hacks (because we are getting [cargo's metadata](https://docs.rs/cargo_metadata/latest/cargo_metadata/diagnostic/struct.Diagnostic.html) instead of [rustc's one](https://doc.rust-lang.org/nightly/nightly-rustc/rustc_errors/diagnostic/struct.Diagnostic.html), so we need to parse it back again). Therefore, I'm think continue to use pre-rendered output and do some matching would be a better idea, like this (pseudo-code)

```rust
fn add_suggestion_actions(pre_rendered: &str, diagnostic: cargo_metadata::diagnostic::Diagnostic, actions: &mut Vec<Action>) {
    if let Some(position) = pre_rendered.find(diagnostic.suggested_replacement) {
        actions.push(Action(position, ActionType::Replacement(diagnostic.suggested_replacement)));
    }
}
fn add_filejump_actions(pre_rendered: &str, diagnostic: cargo_metadata::diagnostic::Diagnostic, actions: &mut Vec<Actions>) {
    let file_loc = format!("{}:{}:{}", diagnostic.file_name, diagnostic.line_start, diagnostic.column_start);
    if let Some(position) = pre_rendered.find(file_loc) {
        actions.push(Action(position, ActionType::Jumpto(diagnostic.line_start, diagnostic.column_start)));
    }
}
```

PR #777, opened 2022-02-08
Add a parallel Axum server

shepmaster:


ui/src/server_axum.rs
```diff
@@ -0,0 +1,511 @@
...
+        .route("/miri", post(miri))
+        .route("/macro-expansion", post(macro_expansion))
+        .route("/meta/crates", get_or_post(meta_crates))
+        .route("/meta/version/stable", get_or_post(meta_version_stable))
+        .route("/meta/version/beta", get_or_post(meta_version_beta))
+        .route("/meta/version/nightly", get_or_post(meta_version_nightly))
+        .route("/meta/version/rustfmt", get_or_post(meta_version_rustfmt))
+        .route("/meta/version/clippy", get_or_post(meta_version_clippy))
+        .route("/meta/version/miri", get_or_post(meta_version_miri))
+        .route("/meta/gist/", post(meta_gist_create))
```
davidpdrsn, line 69: Is the trailing slash intended here? None of the other routes have it.

If someone requests `/meta/gist` axum will send a redirect to `/meta/gist/`.

ui/src/server_axum.rs
```diff
@@ -0,0 +1,511 @@
...
+        let mut parts = rewritten_uri.into_parts();
+        parts.path_and_query = Some(PathAndQuery::from_static("/index.html"));
+        *uri = Uri::from_parts(parts).unwrap();
+    }
+    next.run(req).await
+}
+
+// This is a backwards compatibilty shim. The Rust documentation uses
+// this to run code in place.
+async fn evaluate(Json(req): Json<EvaluateRequest>) -> Result<Json<EvaluateResponse>> {
```
davidpdrsn, line 132: Do you care about the response sent back if the `Json` extractor rejects the request? [These](https://docs.rs/axum/latest/axum/extract/rejection/enum.JsonRejection.html) are the reasons that can happen. It'll return a plain text response in that case.

Customizing these "rejection responses" is done [like this](https://docs.rs/axum/latest/axum/extract/index.html#customizing-extractor-responses). Its a bit more boilerplate that I'd like but its the best solution we have thus far.

davidpdrsn (approved):
Looks good to me. I just had two minor questions.

Goes without saying that I'm pretty excited about this 😊

ui/src/server_axum.rs
shepmaster: Ah, that can probably be changed back now. That was how I found https://github.com/tokio-rs/axum/issues/681

ui/src/server_axum.rs
shepmaster: Hmm. I'll have to investigate what the current behavior is. IIRC, I recover and print out a JSON-formatted error.

davidpdrsn (approved):

Issue #446, opened 2018-12-11
Add a way to pass custom rustc flags


oli-obk:
The playground is often used for experimenting with features and/or creating minimal repros for feature requests and bugs. It would be great if we could make full use of the compiler by also being able to specify arbitrary flags.

My personal motivation is [-Zunleash-the-miri-inside-of-you](https://github.com/rust-lang/rust/pull/56123#issuecomment-440961138), which takes the brakes off the const evaluator and allows demonstrating why we are doing some things the way we are (because other ways are unsound or incoherent or ...).

shepmaster:
I think that @lqd might have had a prototype for this at some point. I remember that we had some light concerns around any options that could do "bad things" to the running process, but there's *so many options* and people keep adding them (eyes @oli-obk suspiciously) that it's difficult to know what works.

It would also be nice to have a list of them and a sentence of "documentation". If you have to open the terminal to figure out the exact name, it's a lot less useful. These can be used to populate an autocomplete/helper like thing.

Notably, any rustflags will also need to be added to the URL for sharing purposes.

RalfJung:
I was imagining that flags could be controlled by a comment in the file, similar to ui tests. Like, if the first line is
```
// compile-flags: FLAGS
```
then those will be passed to rustc/miri/...

> I think that @lqd might have had a prototype for this at some point. I remember that we had some light concerns around any options that could do "bad things" to the running process, but there's so many options and people keep adding them (eyes @oli-obk suspiciously) that it's difficult to know what works.

Given that playground already runs *arbitrary code put in my the user*, I cannot imagine anything beyond that that flags could achieve. You have to have it all sandboxed anyway.

saethlin:
I'm hacking together a prototype implementation of this. It works pretty well to set `MIRIFLAGS` by just adding a `--env` to the Docker invocation.

But when I try to set `RUSTFLAGS`, a build command attempts to recompile the `playground` crate which has an incredible number of dependencies (which is the point). Tossing in `PLAYGROUND_NO_DEPENDENCIES=true` when I set `RUSTFLAGS` works, but seems pretty heavy-handed.

Does anyone have better ideas? @shepmaster ?

Issue #698, opened 2021-04-03
Support running `cargo miri` in situations where target != host (e.g. "cross-interpretation")


thomcc:
The playground has the ability to run code under miri, but it would nice to be able to choose other targets.

A somewhat unknown (but highly useful) feature of miri is that it can run as a cross-interpreter. That is, even on x86_64-apple-darwin (or whatever), if I have, say the `powerpc-unknown-linux-gnu` target installed, I can do:

```
cargo miri test --target=powerpc-unknown-linux-gnu
```

and it will simulate running the test on that architecture, including emulating 32-bit big-endian linux (well, the set of linux syscalls that miri supports, anyway).

Right now I can't do this on the playground, and it would be useful to do quick checks and examples of big-endian code.

---

The downside is needing UI for this, and needing to `rustup target add` on the playground servers. So I understand some hesitancy.

Personally, I think it's most valuable for targets that are very different than x86_64 — e.g. I feel like just saying having an option for big-endian 32 and 64 bit targets would be good enough.

(It would be nice to also have one for target_pointer_width="16", like `msp430-*` or `avr-*`, but these being no_std only, it's hard enough to actually run the code that users probably should just do it locally)

---

Anyway, I thought I'd mention it. I kind of suspect it's the kind of "that'd be nice but doing it in a non-hacky way is likely not worth the effort" thing, but figured I'd suggest anyway. No worries if you'd rather wontfix this.

Note that this is related to https://github.com/integer32llc/rust-playground/issues/446, but not the same, as --target isn't a RUSTFLAG (or a MIRIFLAG), its a flag for cargo.

thomcc:
Hm, I guess I did something wrong here, since it says I added wontfix. I didn't mean to.

If it's just github being weird, can I get some justification there?

shepmaster:
> I missed the small link

Yeah, you'd be surprised at how many people don't read the template before opening an issue.

shepmaster:
> The downside is needing UI for this, and needing to `rustup target add` on the playground servers

Does `cargo miri setup` need to be run for each target?

RalfJung:
> if I have, say the powerpc-unknown-linux-gnu target installed

You don't even need to have the target installed.

> Does cargo miri setup need to be run for each target?

Yes (but this also happens automatically when needed).

RalfJung:
> Note that this is related to #446, but not the same, as --target isn't a RUSTFLAG (or a MIRIFLAG), its a flag for cargo.

[...]

PR #781, opened 2022-02-23
Add a magic comment syntax to set MIRIFLAGS

saethlin:
Based on https://github.com/integer32llc/rust-playground/issues/446#issuecomment-670917927

I really want to be able to set `MIRIFLAGS` when running code in the playground. The community Discord uses a bot that accesses the playground and we use it for demos. It's a real bummer that any time `-Zmiri-tag-raw-pointers` is required to demonstrate something, we can't use the bot. I don't think changing the defaults is a particularly good option either, because we may want to make a point about the impact of `-Zmiri-tag-raw-pointers`. And I recently added `-Zmiri-backtrace` which should be left as its default, but in a pinch can remove backtraces entirely to make the output fit in Discord message length limit.

### DIFF
--- a/ui/src/sandbox.rs
+++ b/ui/src/sandbox.rs
@@ -1,3 +1,4 @@
+use regex::Regex;
 use serde_derive::Deserialize;
 use snafu::{ResultExt, Snafu};
 use std::{ffi::OsStr, fmt, io, os::unix::fs::PermissionsExt, string, time::Duration};
@@ -280,6 +281,27 @@ fn set_execution_environment(
     cmd.apply_backtrace(&req);
 }
 
+fn parse_magic_comments(code: &str) -> Vec<(&'static str, String)> {
+    lazy_static::lazy_static! {
+        pub static ref MIRIFLAGS: Regex = Regex::new(r#"(///|//!|//)\s*MIRIFLAGS\s*=\s*(.*)"#).unwrap();
+    }
+
+    fn match_of(line: &str, pat: &Regex) -> Option<String> {
+        pat.captures(line.trim())
+            .and_then(|caps| caps.get(2))
+            .map(|mat| mat.as_str().trim_matches(&['\'', '"'][..]).to_string())
+    }
+
+    let mut settings = Vec::new();
+    if let Some(first_line) = code.trim().lines().next() {
+        if let Some(miri_flags) = match_of(first_line, &*MIRIFLAGS) {
+            settings.push(("MIRIFLAGS", miri_flags));
+        }
+    }
+
+    settings
+}
+
 pub mod fut {
     use snafu::prelude::*;
     use std::{
@@ -292,9 +314,9 @@ pub mod fut {
     use tokio::{fs, process::Command, time};
 
     use super::{
-        basic_secure_docker_command, build_execution_command, set_execution_environment,
-        vec_to_str, wide_open_permissions, BacktraceRequest, Channel, ClippyRequest,
-        ClippyResponse, CompileRequest, CompileResponse, CompileTarget,
+        basic_secure_docker_command, build_execution_command, parse_magic_comments,
+        set_execution_environment, vec_to_str, wide_open_permissions, BacktraceRequest, Channel,
+        ClippyRequest, ClippyResponse, CompileRequest, CompileResponse, CompileTarget,
         CompilerExecutionTimedOutSnafu, CrateInformation, CrateInformationInner, CrateType,
         CrateTypeRequest, DemangleAssembly, DockerCommandExt, EditionRequest, ExecuteRequest,
         ExecuteResponse, FormatRequest, FormatResponse, MacroExpansionRequest,
@@ -454,7 +476,9 @@ pub mod fut {
 
         pub async fn miri(&self, req: &MiriRequest) -> Result<MiriResponse> {
             self.write_source_code(&req.code).await?;
-            let command = self.miri_command(req);
+
+            let settings = parse_magic_comments(&req.code);
+            let command = self.miri_command(settings, req);
 
             let output = run_command_with_timeout(command).await?;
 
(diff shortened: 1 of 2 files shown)


thomcc:
Related, but probably is more difficult to handle (and can't be bolted on here): https://github.com/integer32llc/rust-playground/issues/698

ui/src/sandbox.rs
```diff
@@ -651,10 +675,23 @@ pub mod fut {
...
+            &self,
+            env_settings: Vec<(&'static str, String)>,
+            req: impl EditionRequest,
+        ) -> Command {
             let mut cmd = self.docker_command(None);
             cmd.apply_edition(req);
 
+            let miri_env =
+                if let Some((_, flags)) = env_settings.iter().find(|(k, _)| k == &"MIRIFLAGS") {
+                    format!("MIRIFLAGS={} -Zmiri-disable-isolation", flags)
```
RalfJung, line 688: Changing the flags like this seems potentially confusing.

If people can set their own flag I am not sure if we even still need the default flag?

saethlin: That's a good point.

I'm concerned that if this flag is removed by default, that people will find programs that used to work and don't anymore. How would we direct them to add the magic comment? I can definitely add something to the help page, but is that enough?

thomcc: Stuff already breaks on the playground over time, as crate versions change. I especially hope there's nobody relying on miri-in-playground not changing in any real way...

RalfJung: If you want to be really fancy, when Miri shows the error saying "pass -Zmiri-disable-isolation to run this operation", the playground could turn that into a link that adds the relevant magic command. ;) The playground has such links for other situations. Those are easier though since here, if there already is a MIRIFLAGS comment, one has to now *change* the existing comment...

saethlin: I could probably handle most cases. Would it be better to do this edit which is almost always correct, or append something to the output that points to documentation? Both options seem sort of bad in slight ways. I like that currently the playground _just_ echoes cargo output.

ui/src/sandbox.rs
```diff
@@ -280,6 +281,27 @@ fn set_execution_environment(
     cmd.apply_backtrace(&req);
 }
 
+fn parse_magic_comments(code: &str) -> Vec<(&'static str, String)> {
+    lazy_static::lazy_static! {
+        pub static ref MIRIFLAGS: Regex = Regex::new(r#"(///|//!|//)\s*MIRIFLAGS\s*=\s*(.*)"#).unwrap();
```
danielhenrymantilla, line 286: An outer attribute seems weird, here:

```suggestion
        pub static ref MIRIFLAGS: Regex = Regex::new(r#"(//!|//)\s*MIRIFLAGS\s*=\s*(.*)"#).unwrap();
```

ui/src/sandbox.rs
```diff
@@ -280,6 +281,27 @@ fn set_execution_environment(
...
+    }
+
+    fn match_of(line: &str, pat: &Regex) -> Option<String> {
+        pat.captures(line.trim())
+            .and_then(|caps| caps.get(2))
+            .map(|mat| mat.as_str().trim_matches(&['\'', '"'][..]).to_string())
```
danielhenrymantilla, line 292: Hoe does this handle trailing whitespace?

danielhenrymantilla (approved):

ui/src/sandbox.rs
saethlin: Trailing whitespace outside quotes should be trimmed off, trailing whitespace inside quotes is currently retained I think. Is that a problem?

ui/src/sandbox.rs
saethlin:
```rust
/// MIRIFLAGS=-Zmiri-tag-raw-pointers
fn main() {
}
```
seems plausible to me. My instinct here is to err on the side of accepting as much as possible, because I think the only alternative we have is silently ignoring things.

RalfJung: I agree... whenever I use magic comments on ui tests I am nervous that they might be silently ignored.

ui/src/sandbox.rs
danielhenrymantilla: No that's fine, I missed the `line.trim()` input beforehand 👍

ui/src/sandbox.rs
danielhenrymantilla: So, the thing is that:

```rs
#![feature(…)]

/// MIRIFLAGS=-Zmiri-tag-raw-pointers
fn main() {
```

would fail, IIUC, since it wouldn't be the first-line of a file. And I can imagine the transition from your snippet to mine occurring quite easily, especially given that

```rs
/// MIRIFLAGS=-Zmiri-tag-raw-pointers
#![feature(…)]

fn main() {
```

would not be valid Rust.

I'd rather have Miri start saying with which `MIRIFLAGS` it is being run (which seems like a nice addition overall), rather than have an outer attribute be accepter where an inner one should be used: it is all too reminiscent of:

```rs
//! src/lib.rs
#[forbid(unsafe_code)]

use …;

fn foo… /* actually allowed to use `unsafe`! */
```

RalfJung: > would fail, IIUC, since it wouldn't be the first-line of a file.

Hm... good point. Playground will also add `feature` lines at the very top of the snippet when you click the error message that complains about a missing feature gate.

saethlin: I think `#![feature(...)]` is pretty problematic here. I was afraid magic comments would run into this kind of complexity. 

Parsing the magic flag on lines other than the first makes some sense in response to this, but then we run into this:
```rust
// This example works totally fine by default. But it breaks if we set
// MIRIFLAGS=-Zmiri-tag-raw-pointers
fn main() {
    🤦 
}
```

I agree that Miri should report its configuration somewhere.

Other things we could do...
1. Add a new UI element. I don't really want to do this because it's a lot of work, and also MIRIFLAGS is quite unique in that it needs to be provided as an environment variable when `cargo-miri` is run. All other build options we would want as arguments to `cargo rustc`.
2. Try to parse the first comment line without leading whitespace. Is it sufficiently unlikely to accidentally set MIRIFLAGS this way?

thomcc: I don't think this is unreasonable. It's not like people are shipping code using the playground, and this seems to be quite an edge case (https://grep.app/search?q=//%20MIRIFLAGS and https://cs.github.com/?scopeName=All+repos&scope=&q=%22%2F%2F+MIRIFLAGS%22 say enough of one that it doesn't exist in open source rust code).

saethlin: @thomcc Can you clarify what exactly "it" is here?
> I don't think this is unreasonable.

thomcc: > Parsing the magic comment on lines other than the first makes some sense in response to this, but then we run into this...

The issue you are describing here.

thomcc: I think it's fine if that turns on that MIRIFLAG

Nilstrieb:
Any updates on this?

PR #796, opened 2022-04-14
Enable all stable smallvec features

rodrimati1992:


shepmaster:
Smallvec themselves needs to enable the features they think would be in the playground. See https://github.com/integer32llc/rust-playground/issues/192#issuecomment-388589736. Once they have published a new release with that metadata, it will be automatically included at the next update.

PR #803, opened 2022-05-18
Allow third parties to easily deploy custom playgrounds

rickwebiii:
# Summary
This PR includes a number of improvements that make deploying your own playground significantly easier. It allows you to better customize which crates you allow. It further features a deployment executable that manages deploying a self-contained playground on EC2 serving content over HTTP. This eliminates all manual setup steps for instances.

## Customizations
* Can limit the number of top crates to include in your playground (set to zero to include include *no* top crates).
* Can omit cookbook crates
* Combining the previous 2 options and tweaking crate-modifications.toml, you can deploy a playground that serves only the crates in your organization.

## Automated deployment
Our deployment architecture looks like the following:
### Manual one-time setup
* Route53 hosted zone with an "A" DNS record pointed to...
* ... an API Gateway. This API gateway has CORS enabled (as the playground does), uses a an ACM certificate for HTTPS
with our custom domain name. Our config simply proxies all HTTP requests to...
* ... an elastic IP.

### Automatic deployment
1. Compile ui, frontend, deployment app, rust-playground-top-crates 
2. Run rust-playground-top-crates, build docker images and publish them
3. Spin up and tag an EC2 instance and install prereqs on the machine with a user-data script.
3. Copy ui, frontend to the EC2 instance.
4. run fetch.sh on the new instance
5. systemctl start playground on the EC2 instance.
6. Assign the elastic IP to the new instance
7. terminate all tagged instances that aren't the new instance.

This approach has a number if differences with today's model:
* No cron jobs to sync images. Images get downloaded one time on machine setup. To update the crates, run another deployment.
* No manual setup for EC2 instances
* AWS manages the HTTPS certificate in ACM, so you don't have to run a certbot cron job to renew it.

### Binaries build under a workspace
`ui`, `deployment`, and `top-crates` now reside under a `[workspace]`. This improves build times as dependencies are shared and reused across components.

This is a dump of what we have working on our end, so there's probably a bit more work to migrating the official playground to using this model.


compiler/base/crate-information.json
```diff
@@ -1,24 +1,4 @@
 [
```
rickwebiii, line 1: This file can be reverted.

deployment/aws_launch_instance.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
```
rickwebiii, line 1: This file can be deleted. The deployment app supercedes it.

top-crates/crate-modifications.toml
```diff
@@ -3,6 +3,5 @@ exclusions = []
 additions = [
     # Crates that the compiler recommends
     # rg 'https://crates.io/crates/[a-z_-]+' --glob '*.stderr' --only-matching --no-line-number --no-filename | sort | uniq | cut -d '/' -f 5
-    "async-trait",
-    "async-recursion",
+		"sunscreen",
```
rickwebiii, line 6: This can be reverted.

deployment/src/main.rs
```diff
@@ -0,0 +1,489 @@
...
+        &["./fetch.sh", &docker_repository],
+    )?;
+
+    run_ssh_command(
+        &ip_name,
+        &key_file,
+        &["sudo", "systemctl", "start", "playground"],
+    )?;
+
+    std::fs::remove_file(&key_file)?;
```
rickwebiii, line 471: Should add a rollback mechanism to terminate the new instance if setting up the instance fails after creating it, but before the the IP association.

shepmaster:
I have not read through this at all, just some high-level comments.

1. I'm sympathetic to the general goal presented here.
2. I'd much rather use the (relatively new) [Ansible playbooks](https://github.com/rust-lang/simpleinfra/blob/master/ansible/playbooks/playground.yml) instead of having a completely disjoint deploy mechanism.

rickwebiii:
Closing this request until I've reworked the deployment situation...

PR #816, opened 2022-06-22
Make PLAYGROUND_GITHUB_TOKEN optional

dtolnay:
I am working on a PR for something totally unrelated to gists, and didn't want to bother setting up a GitHub token just so that the server wouldn't crash on startup.

PR #825, opened 2022-07-03
Speed up base build and shrink image

dtolnay:
This is a substantial refactor of `top-crates` (but split up into easy commits for review).

Basically the playground was previously redundantly building multiple redundant versions of lots of crates redundantly. In the case of the following dependency graph:

- `playground`
  - normal dependency on `A`
    - normal dependency on `C` with no features
  - dependency on `B` (proc macro)
    - build dependency on `C` with feature `x`

cargo would build 2 copies of `C`, with and without the feature `x` enabled, because one is for the "target" and one is for the "host", even though in the context of the playground the "target" and "host" are the same architecture and one shared copy of `C` would have been fine.

If `C` is somewhere way deeper in the dependency graph (like `proc-macro2`, which was affected by this) then *everything* downstream of it would potentially be built multiple times.

The way to fix this is to ensure that every transitive dependency which appears as both a "target" dependency and as a "host" dependency (dependency of a proc macro or build script) has the same feature selected in both cases. Then cargo will build it just once, since the target and host are set to the same architecture.

To do this, we must run the cargo resolver multiple times in a loop, copying the selected features in the output of each run onto the input of the next run. This way if cargo resolves a dependency on feature `x` of some crate on just the "host" or just the "target", the next resolver will proceed with `x` enabled on *both* "host" and "target". Eventually the resolver will reach a fixed point because the process is additive; we only ever *add* features from one run to the next. In practice this takes 3 resolver runs today.

### Before

- `time cargo build`: real 0m37.935s user 11m48.885s
- `du -s target`: 1.91 GB

### After

- `time cargo build`: real 0m35.595s user 7m7.891s
- `du -s target`: 1.59 GB

This is 40% less CPU time and 17% smaller Docker image.

### Rationale

As you might guess, I don't care about the build time or image size in isolation. The real reason for this change is that I need this in order to implement support for `#![crate_type = "proc-macro"]` in the playground.

When the playground runs a build with `[lib] proc-macro = false` (the default), everything under `[dependencies]` is considered a "target" dependency. With `[lib] proc-macro = true`, everything under `[dependencies]` becomes a "host" dependency. If we don't ensure all transitive dependencies have the same features enabled in both modes, then `#![crate_type = "proc-macro"]` can't work because the precompiled deps won't apply; cargo will start rebuilding a bunch of deps with a different set of features than the prebuilt ones have, and the playground build will inevitably time out after 10 seconds.

### DIFF
--- a/top-crates/src/main.rs
+++ b/top-crates/src/main.rs
@@ -27,12 +27,21 @@ struct TomlPackage {
     resolver: String,
 }
 
+/// Profile used for build dependencies (build scripts, proc macros, and their
+/// dependencies).
+#[derive(Serialize)]
+#[serde(rename_all = "kebab-case")]
+struct BuildOverride {
+    codegen_units: u32,
+}
+
 /// A profile section in a Cargo.toml file
 #[derive(Serialize)]
 #[serde(rename_all = "kebab-case")]
 struct Profile {
     codegen_units: u32,
     incremental: bool,
+    build_override: BuildOverride,
 }
 
 /// Available profile types
@@ -67,10 +76,12 @@ fn main() {
             dev: Profile {
                 codegen_units: 1,
                 incremental: false,
+                build_override: BuildOverride { codegen_units: 1 },
             },
             release: Profile {
                 codegen_units: 1,
                 incremental: false,
+                build_override: BuildOverride { codegen_units: 1 },
             },
         },
         dependencies,
--- a/top-crates/src/main.rs
+++ b/top-crates/src/main.rs
@@ -16,6 +16,8 @@ struct TomlManifest {
     profile: Profiles,
     #[serde(serialize_with = "toml::ser::tables_last")]
     dependencies: BTreeMap<String, DependencySpec>,
+    #[serde(serialize_with = "toml::ser::tables_last")]
+    build_dependencies: BTreeMap<String, DependencySpec>,
 }
 
 /// Header of Cargo.toml file.
@@ -84,7 +86,8 @@ fn main() {
                 build_override: BuildOverride { codegen_units: 1 },
             },
         },
-        dependencies,
+        dependencies: dependencies.clone(),
+        build_dependencies: dependencies,
     };
 
     // Write manifest file.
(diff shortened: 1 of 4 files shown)


dtolnay:
I promise it was a coincidence that this opened literally the same minute that you merged #816 &mdash; I have been working on it for several hours.

I would be happy to split this up over multiple PRs if that would make the review easier. I didn't want to just start at the beginning and start sending small PRs that wouldn't necessarily make sense without seeing what this adds up to.

top-crates/src/main.rs
```diff
@@ -53,7 +54,24 @@ fn main() {
     let modifications: Modifications =
         toml::from_slice(&d).expect("unable to parse crate modifications file");
 
-    let (dependencies, infos) = rust_playground_top_crates::generate_info(&modifications);
```
shepmaster, line 56: This is the only part that I have qualms about. Niko and team were [doing some analysis](https://github.com/optopodi/optopodi) and requested that this code be made available so that they could get (more-or-less) the same list as the playground.

I don't mind changing the public API at all, but it would probably be nice if we kept a simpler API for them.

shepmaster (commented):
Everything seems fine, and special thanks for taking the time to improve some of the code you came across. I've mostly been making minimal changes to this as new Cargo versions are released and haven't necessarily been trying to have the best code ever.

top-crates/src/main.rs
dtolnay: Ah okay, did not know. The only reason I pulled `generate_info` out so that the loop lives in main.rs instead of lib.rs was so that I could put `println!("resolve #{i}")` in there. It felt weird for the println to be in the lib, and main was already printing things using println.

Would it be okay to move the big green block below this (`let config` through `let infos`) back into lib.rs as `generate_info`, even with the println? Or take out the println? Or keep it here, but put another copy without println into lib.rs for Niko and friends? Something else?

dtolnay: I just removed b56e90a6ec3eb1315bbe75d606683c67d33f3810 entirely from the PR. This runs fast enough that progress reporting isn't important.

shepmaster:
Thanks!

PR #826, opened 2022-07-03
Support `#![crate_type = "proc-macro"]`

dtolnay:
> **Warning**
> **This doesn't work until after [#825](https://github.com/integer32llc/rust-playground/pull/825) has been merged.**

Screenshot from 2022-07-02 17-58-05


shepmaster:
Could you link me the code you ran there so I could add it to the test suite?

dtolnay:
https://gist.github.com/rust-play/23b7c1923796dce51412f4b5e54852c7

shepmaster:
When I remove the `#[test]`, I see

> warning: library `playground` should only specify `proc-macro = true` instead of setting `crate-type`

Is that worrisome?

dtolnay:
I do not get that. Are you sure you are running against a container that contains the _modify-cargo-toml_ change from this PR?

Full output with `#[test]`:

```console
Standard Error

   Compiling playground v0.0.1 (/playground)
    Finished test [unoptimized + debuginfo] target(s) in 0.69s
     Running unittests src/lib.rs (target/debug/deps/playground-cc6c5b427caf9369)
   Doc-tests playground
error: test failed, to rerun pass '--doc'

Standard Output


running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s


running 1 test
test src/lib.rs - demo (line 6) ... FAILED

failures:

---- src/lib.rs - demo (line 6) stdout ----
wow wow
error[E0412]: cannot find type `aaa` in this scope
 --> src/lib.rs:8:10
  |
4 | playground::demo!();
  | ------------------- similarly named struct `Aaa` defined here
5 | type T = aaa;
  |          ^^^ help: a struct with a similar name exists: `Aaa`

error: aborting due to previous error

For more information about this error, try `rustc --explain E0412`.
Couldn't compile the test.

failures:
    src/lib.rs - demo (line 6)

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.24s
```

Full output without `#[test]`:

```console
Standard Error

   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.58s

Standard Output

Warnings

No main function was detected, so your code was compiled
but not run. If you’d like to execute your code, please
add a main function.
```

shepmaster:
I did rebuild the containers, but I'll double check that during my next attempt.

I am getting the (expected?) output:

```
---- src/lib.rs - demo (line 6) stdout ----
wow wow
error[E0412]: cannot find type `aaa` in this scope
 --> src/lib.rs:8:10
  |
4 | playground::demo!();
  | ------------------- similarly named struct `Aaa` defined here
5 | type T = aaa;
  |          ^^^ help: a struct with a similar name exists: `Aaa`
```

Is a doc test the only way to exercise the proc macro with the current playground setup?

If so, we might want to add a follow-up issue to try and regex for <code>/// ```</code>-like lines and enable test mode for that. 

I suppose somehow supporting multiple crates would be another solution...

dtolnay:
> Is a doc test the only way to exercise the proc macro with the current playground setup?

That's right. An ordinary #\[test\] unit test will compile and run as well, but does not **depend on** the proc macro, so it can't use the macro as a macro, unlike an integration test or doc test. At best it can test helper functions inside of the implementation.

PR #854, opened 2022-10-07
fix miri sysroot location

RalfJung:
The miri sysroot moved, so this hard-coded location needs to be updated.


shepmaster (approved):

shepmaster:
Heh rot

Issue #638, opened 2020-10-06
main() detection doesn't catch a manually defined `extern "C" fn main()`


abonander:
Playground doesn't detect that the following code has a `main()` function:

```rust
#![no_main]

use std::os::raw::*;

#[no_mangle]
pub extern "C" fn main(argc: c_int, argv: *const *mut c_char) -> c_int {
    println!("Hello from main({}, {:p})", argc, argv);
    0
}
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=736713393bf0a8b9cb307dcc5a46fdc4

This is a pretty niche edge case, so feel free to close this if you don't think it's worth the effort. I didn't see anything mentioning this specific case in the issue tracker so I figured I'd report it.

Issue #624, opened 2020-07-28
`main` function definition not own its own line is not detected


MonliH:
After entering the following code:
```rust
use std; fn main(){ println!("Hello, world!"); }
```
and clicking the "Build" button, I get the following message:
```
No main function was detected, so your code was compiled
but not run. If you’d like to execute your code, please
add a main function.
```
No code in the main function is run.

<br>
<br>

After a bit more testing, I've found that if the `main` function's signature is not on its own line, the playground fails to detect the function:
```rust
// Also doesn't detect main
const foo: i32 = 10; fn main(){}

// Also doesn't detect main
use std; fn main(){
}

// Does detect main
use std;
fn main() {}
```

sollyucko:
Detection code, AFAICT: https://github.com/integer32llc/rust-playground/blob/master/ui/frontend/selectors/index.ts#L13

shepmaster:
I don’t really consider this to be a problem. Why would you *choose* to write nonstandard Rust? You can run rustfmt to fix it. If you don’t, you can use the drop down and choose run manually.

MonliH:
I came across this because I was playing around with **code golfing** in rust :smile:. I needed to store the code somewhere so other people could run it and see my solution, so that's why. 

I know its sort of a niche use case, although fixing small things like this makes for a more polished project. That being said, I can see why this wouldn't be high priority. I might be able to patch this as well.

sollyucko:
I think removing the `^` at the start that requires the RegExp to match at the start of the line should probably work. The starting `\s*` (matches optional whitespace) and the ending `m` (`^` and `$` can match the start/end of the line instead of the whole string) can optionally be removed.

shepmaster:
> removing the ^ at the start

```
// fn main() {}
```

MonliH:
Hmm, yeah. Seems like there would be a a few more edge cases you need to deal with, if you remove that:
```rust
/* fn main() {} */
```
```rust
const FOO: &'static str = r#"fn main() {}
"#;
```

Maybe look for a main function on the server side instead? i.e. is there a way to detect if a main function exists after compiling it?

shepmaster:
> after compiling it?

The compilation command and even the file name itself changes based on how it's being compiled.

sollyucko:
I guess you could build it, then analyze the binary to detect the presence of a `main` function (Run) or a test runner (Test), then determine what to do with it.

Also, multi-line comments can still trick the selector:

```
/*
[...]

Issue #511, opened 2019-07-08
Nested `main` causes build error


joshlf:
The following code snippets seem to cause the playground to interpret the file as a binary (including changing the "build" button to say "run"), but then `rustc` can't find a crate-level `main` function, and so fails compilation:

```rust
mod foo {
    fn main() {}
}
```

```rust
fn foo() {
    fn main() {}
}
```

IMO, only `main` functions at the crate level should cause this behavior.

shepmaster:
Yep!

The automatic selection is done [via simple regexes](https://github.com/integer32llc/rust-playground/blob/5189288f6d67c2d5c11e6153f81a034d058ceefd/ui/frontend/selectors/index.ts#L8-L40). 

If you’d like to fix this, you are welcome to explore an alternate regex or parsing mechanism to make the decision.

In the meantime, you can always override the automatically-selected mode:

![image](https://user-images.githubusercontent.com/174509/60851922-93178e80-a1c3-11e9-9adc-a5134013fdea.png)

joshlf:
Ah, seems like a much more involved fix. Maybe eventually we'll want to wasm-ify the Rust parser, but I think for the time being this is probably a small enough wart that it can be ignored. Thanks for the quick reply!

shepmaster:
> wasm-ify the Rust parser

It’s possible (see #357), but one important thing to continue to preserve is the ability to handle malformed Rust code.

Issue #503, opened 2019-06-13
does not support  #![feature(main)]


matthiaskrgr:
````rust
#![feature(main)]

#[main]
fn a() {
    println!("hi");
}
````
When I run it locally, I get a "hi", but playground says

````
   Compiling playground v0.0.1 (/playground)
warning: function is never used: `a`
 --> src/lib.rs:6:1
  |
6 | fn a() {
  | ^^^^^^
  |
  = note: #[warn(dead_code)] on by default

    Finished dev [unoptimized + debuginfo] target(s) in 1.43s

Warnings

No main function was detected, so your code was compiled
but not run. If you'd like to execute your code, please
add a main function.
````

https://play.rust-lang.org/?version=nightly&mode=debug&edition=2018&gist=7ff692f07c7c3562df289936dbc97eda

shepmaster:
Yes, the automatic detection of whether to run `cargo run` / `cargo build` / `cargo test` is based on a simple regex match. You don't have `fn main()` so the regex isn't triggered. You need to expressly request to run the code:

![image](https://user-images.githubusercontent.com/174509/59435525-d09f1c80-8dbb-11e9-91bf-99d30579add9.png)

You'll note that the button says "build" not "run" before you do this.

matthiaskrgr:
Thanks for clarifying! :+1:

PR #863, opened 2022-11-10
Add detection of main under more circumstances

c-git:
Foreword: I tried my best to not affect the runtime performance of the regex matching but I was only able to account for the cases I could understand / think of. I've listed those below and how I tried to mitigate them.

The cases of main detection that are intended to be added by this PR are:

- Code with no comments but ends in a semi colon before the declaration of main. No comments was required to prevent `//` being included to comment out the line. This was done by not allowing `/` as part of the code before the `;` While this is more restrictive than necessary it didn't seem likely that the code before the main would include `/` so didn't seem worthwhile to take the performance hit to make the check less restrictive. To prevent long regex scans going multiple lines I also disallowed newlines in the code by including `\n` and `\r` as disallowed characters. (The whole point is code on the same line as main so this seemed fine and would help prevent unnecessarily matching all lines before main when it doesn't matter).

- Allow multiline comments within the arguments area of main that both start and end within the brackets. Not expecting a substantial performance cost from this one because this case is near the end of the matching.

I used the following test cases:

```rust
fn main() { // Should work
    println!("Hello, world!");
}

//;fn main() { //Shouldn't work
    println!("Hello, world!");
}


use std; fn main(){ println!("Hello, world!"); } // Should work

const fn main() { // Should work
    println!("Hello, world!");
}

/* fn main() {} */ // Shouldn't work

fn main(/* comment */) { // Should work
   // snip
}
```


adwinwhite:
It's really nice to have detailed explanation of the regex change.

The small caveat in your PR,
> This was done by not allowing / as part of the code before the ; While this is more restrictive than necessary it didn't seem likely that the code before the main would include / so didn't seem worthwhile to take the performance hit to make the check less restrictive.

While it's indeed uncommon to write `/` before `main()`, it will confuse people greatly if happens (In code golf like #624).

I wonder if it's possible to solve the problem of `main()` in comments **once for all**.
One approach would be **filter out all comments** before detecting `main()`.
Since the editor detects comments in real-time, the runtime cost shouldn't high and it's only executed when the user runs/builds.

I just did a small benckmark with following code,
```javascript
const SINGLELINE_COMMENTS_RE = /\/\/.*$/gm;
const BLOCK_COMMENTS_RE = /\/\*.*\*\//gs;

const url = "https://raw.githubusercontent.com/rust-lang/rust/master/library/alloc/src/collections/btree/map.rs";
fetch(url).then(res => res.text())
            .then(text => {
                const code = text;
                console.log("length of code: ", code.length);
                const startTime = performance.now();
                const code_without_comments = code.replaceAll(SINGLELINE_COMMENTS_RE, "").replaceAll(BLOCK_COMMENTS_RE, "");
                const endTime = performance.now();

                console.log("It takes ", (endTime - startTime) / 1000, "seconds");
                console.log("trimmed code length: ", code_without_comments.length);
            });
```
and got result(repeated, it usually cost less than 1 millisecond which is barely tangible to user),
```
[LOG]: "length of code: ",  77816 
[LOG]: "It takes ",  0.00040000009536743164,  "seconds" 
[LOG]: "trimmed code length: ",  49181 
```

Without comments, detection of `main()` could be much easier.
And the bug that `main()` might be contained in block comments is eliminated.
New regex:
```javascript
const HAS_MAIN_FUNCTION_RE = /^(.*;)?\s*(pub\s+)?\s*(const\s+)?\s*(async\s+)?\s*fn\s+main\s*\(\s*)?\s*\)/m;
```

- `\n\r` is no longer needed since `.` won't match newline.

<br /><br />
**However** considering all the forms `main()` may appears,

1. [main<>()](https://github.com/integer32llc/rust-playground/pull/669)
2. [pub extern "C" fn main(...)](https://github.com/integer32llc/rust-playground/issues/638)
3. [#main](https://github.com/integer32llc/rust-playground/issues/503)
4. [nested main() in submodule](https://github.com/integer32llc/rust-playground/issues/511)

We'd better just add a [notice](https://github.com/integer32llc/rust-playground/issues/624#issuecomment-666857893) when `main` is not detected.

Current [warning](https://github.com/integer32llc/rust-playground/blob/f604243a9eb56364e58e491e20b86dfbf5b292ab/ui/frontend/Output/Execute.tsx#L34) doesn't guide user to change mode to `RUN`.

<br />
<br />

BTW, you can add the regex tests to [this file](https://github.com/integer32llc/rust-playground/blob/master/ui/frontend/selectors/index.spec.ts) .

c-git:
Thank you for the response. I'm sorry but I'm not quite sure what action is being requested.

adwinwhite:
Sorry for the lengthy words.  

I just realized that detection is done in real-time.
Then comments filtering will affect performance much more.

My overall suggestion is that we'd better add a notice to guide user to change execution mode manually when main() detection fails.

And change of regex should be accompanied by tests.

c-git:
Thanks for explaining,

I have suggested a change to the notice [here](https://github.com/integer32llc/rust-playground/issues/624#issuecomment-1310883071) and I'm not sure where the tests should be added. If you can guide me to them, I'd be happy to add them as they are documented in the comments here already.

shepmaster:
> I wonder if it's possible to solve the problem of `main()` in comments **once for all**.

It's certainly _possible_, but the cost of such a fix is unclear to me. The best answer I can think of would involve running the Rust parser itself (or something equivalent) and having it report if a function named `main` is found.

The current architecture of the playground doesn't have a great way of doing this. Right now, it would involve sending the code to the backend, running something, then getting a response. That's far too slow for something that should roughly correspond to every keystroke.

Another technique would be to compile something into Wasm and execute that in the browser. It would allow us to reuse some authoritative Rust parser without a network connection. However, it would bring in a large new dependency which might not work for all of our browsers. 

I don't think either of these directions is appropriate right now, but are worth considering!

> Since the editor detects comments in real-time

I'm not sure exactly which editor you are referring to here, but if you mean Ace / Monaco, this is another interesting idea I hadn't considered. Those libraries already have some established JS for parsing Rust code (with differing quality). If we could piggyback off of the work that they've already done, we could get the result "for free". However, it's not clear how that would work for the simple editor, or if Ace/Monaco have appropriate APIs to query what we want.

> One approach would be **filter out all comments** before detecting `main()`.

This seems like the best compromise at this point in time. A quick search found [Regex to match a C-style multiline comment](https://stackoverflow.com/q/13014947/155423) and there are a few JS libraries out there that claim to remove comments (although many have lots of open issues). However, a purpose-built JS (or just a file here) that removes Rust comments[^1] may be more appropriate.

A benefit to this approach is that it would also apply to cases beyond just `fn main`, such as:

```rust
/*
#[test]
*/
fn x() {}
```

[^1]: For example, this is **invalid** Rust, but I think most regex solutions would accept it: `/* /* */`

shepmaster:
All that being said, I'm OK with incremental improvements to the current scheme. Your unique test cases _do_ need to be added to the test file (as mentioned above, those are in `selectors/index.spec.ts`). You can run the tests via `yarn test`

c-git:
Ok thanks, will run add them when I get a chance.

Ezrashaw:
Has anyone considered macro-expanded `main` functions?

This is naive but: couldn't this logic be done with the compiler? Then you have an *authoritative* answer as to whether the crate has a main function or not. Trying to parse Rust's grammar with regexs probably isn't going to work in the long run; I'm sure we can come up with failing examples indefinitely.

shepmaster:
> This is naive but: couldn't this logic be done with the compiler?

Please take the time to [read previous comments](https://github.com/rust-lang/rust-playground/pull/863#issuecomment-1387619049).

Ezrashaw:
Gosh, I can't believe I missed that lol. I will note that just the lexing/parsing stage is very fast (especially if it wasn't over the network), `rustc` spends most of its time after it has a list of function names.

c-git:
# TLDR

I don't think I will add the testing myself. Happy to help if someone else wants to take it to the finish line but I won't be able to get it there myself.

# Details
I've really wanted to setup the testing but I haven't been able to make the time to install and become familiar with yarn. And I've been doing some reflection on the things on my todo list and doing a bit of house keeping. Realistically this isn't something I can do quickly so it will never get done in that category and hence it would have to be otherwise productive time that I'd have to allocate to this. And while I really want to help out I don't realistically think I will schedule the time to learn yarn over something else. If someone else wants to take what I've done and bring it to the finish line, I'm happy to help with that but don't think I will get around to doing it myself.

c-git:
I'm sorry I wasn't able to bring this to completion. Thanks for the help.

PR #871, opened 2022-12-25
feat(ui/frontend): add shortcuts for tools

### LINKED ISSUE #844: Add shortcut to run rustfmt
I think it would be nice if one could run `rustfmt` from a keyboard shortcut, similar to `ctrl`+`return` to run code, or `ctrl`+`F9` on compiler-explorer.org.

ShaunSHamilton:
Adds shortcuts to tools. Global event listener is added for the following keydown events:

| Keys | Result |
| ----- | ------- |
| `Ctrl + Alt + f` | Rust Fmt on the code |
| `Ctrl + Alt + m` | Miri |
| `Ctrl + Alt + x` | Macro Expansion |
| `Ctrl + Alt + c` | Clippy |

The above can easily be changed. However, certain combinations are reserved by different browsers.

<details>
  <summary>What this looks like</summary>

  I realise the background colour from the issue was probably for emphasis. So, am obviously happy to change/remove it.

  image
</details>

Closes #844 

---

Aside: More ESLint/editor rules should be added, as I constantly needed to avoid saving with formatting.

### DIFF
--- a/ui/frontend/Playground.tsx
+++ b/ui/frontend/Playground.tsx
@@ -112,10 +112,13 @@ const Playground: React.FC = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, []);
 
-  useKeyDown(['Control', 'Alt', 'f'], handleRustFmt);
-  useKeyDown(['Control', 'Alt', 'c'], handleClippy);
-  useKeyDown(['Control', 'Alt', 'm'], handleMiri);
-  useKeyDown(['Control', 'Alt', 'e'], handleMacroExpansion);
+  const shortcutMap = new Map([
+    [['Control', 'Alt', 'f'], handleRustFmt],
+    [['Control', 'Alt', 'c'], handleClippy],
+    [['Control', 'Alt', 'm'], handleMiri],
+    [['Control', 'Alt', 'x'], handleMacroExpansion],
+  ]);
+  useKeyDown(shortcutMap);
 
   return (
     <>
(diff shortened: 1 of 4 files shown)


tgross35:
Could I suggest adding something like `ctrl+alt+t` to run tests? Helpful when you have both a `main` and `#[test]`s

tgross35:
There's also https://github.com/rust-lang/rust-playground/issues/814 which suggested capturing `ctrl+s` to run the code. That's the current behavior of the monaco editor, but it could be good to capture it for the others

ui/frontend/Playground.tsx
```diff
@@ -88,15 +90,43 @@ const ResizableArea: React.FC = () => {
...
+    dispatch(actions.performMiri());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+  const handleMacroExpansion = useCallback((_event) => {
+    dispatch(actions.performMacroExpansion());
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  const shortcutMap = new Map([
+    [['Control', 'Alt', 'f'], handleRustFmt],
```
ian-h-chamberlain, line 116: FWIW, `shift+alt+f` is the default VSCode shortcut for formatting a document on macOS and Windows (but not Linux, it seems?), so users of the Monaco editor might find that more familiar, although it doesn't use `Control` like the shortcuts propsed here.

On macOS at least (and I would guess on other OSes as well), some of these shortcuts seem to have uses already within the editor:
* `cmd+alt+f` seems to open a find+replace dialog in `ace` and `monaco`
* `cmd+alt+c` toggles case-sensitivity for searches in `monaco`

----

This PR looks like it has been open a while without much movement, but I just wanted to provide some feedback on the choice of shortcuts. If there is any interest in reviving this I think it would definitely be a nice feature to have!

PR #896, opened 2023-02-06
update error-index hyperlinks

dawnofmidnight:
This updates the hyperlinks for error codes to point to the new error index locations (as updated in rust-lang/rust #100922). For example, what used to link to `https://doc.rust-lang.org/stable/error-index.html#E0005` will now link to `https://doc.rust-lang.org/stable/error_codes/E0005.html`. Currently, a redirect makes this work fine, but it's probably best to use the new locations.


shepmaster (approved):

shepmaster:
Thanks!

PR #899, opened 2023-02-17
Optionally perform sandbox execution via the WebSocket

shepmaster:
/cc @adwinwhite


ui/frontend/actions.ts
```diff
@@ -128,7 +130,27 @@ export enum ActionType {
...
+  WebSocketError = 'WEBSOCKET_ERROR',
+  WebSocketConnected = 'WEBSOCKET_CONNECTED',
+  WebSocketDisconnected = 'WEBSOCKET_DISCONNECTED',
+  WebSocketFeatureFlagEnabled = 'WEBSOCKET_FEATURE_FLAG_ENABLED',
+  WSExecuteRequest = 'WS_EXECUTE_REQUEST',
+  WSExecuteResponse = 'WS_EXECUTE_RESPONSE',
+}
+
+const ExecuteExtra = z.object({
+  isAutoBuild: z.boolean(),
```
adwinwhite, line 148: What does `isAutoBuild` do?  Rebuild at every keystroke?

adwinwhite:
Sorry for the delay. 

I read all commits line by line. Seems all good to me. (And learned a lot on async primitives, Mutex use, Redux and json schema)

Do you think there is a case when an unique request id is needed for each request/response? Although I can't think of any for now. There is no strong dependency between requests, and responses are almost guaranteed to arrive in order.

shepmaster:
> I should probably remove that...

I [added a commit](https://github.com/integer32llc/rust-playground/pull/899/commits/955af306f2beedd56391527f586f5d9c2b12b716) removing it.

adwinwhite:
The new commit also looks good.

PR #912, opened 2023-03-21
Wait for a WebSocket connection to close before reconnecting

shepmaster:
As identified by @aguspdana, the original implementation attempted to reconnect every so often, even if other connection attempts were still in progress! This could lead to a situation where multiple WebSockets were connected, which could explain the recent surge in connections we saw *after* the incident where the server could not respond.

### DIFF
--- a/ui/frontend/reducers/websocket.ts
+++ b/ui/frontend/reducers/websocket.ts
@@ -1,7 +1,7 @@
 import { PayloadAction, createSlice } from '@reduxjs/toolkit';
 import z from 'zod';
 
-import { createWebsocketResponseSchema } from '../websocketActions';
+import { createWebsocketResponseSchema, makeWebSocketMeta } from '../websocketActions';
 
 export type State = {
   connected: boolean;
@@ -14,6 +14,11 @@ const initialState: State = {
   featureFlagEnabled: false,
 };
 
+const websocketConnectedPayloadSchema = z.object({
+  iAcceptThisIsAnUnsupportedApi: z.boolean(),
+});
+type websocketConnectedPayload = z.infer<typeof websocketConnectedPayloadSchema>;
+
 const websocketErrorPayloadSchema = z.object({
   error: z.string(),
 });
@@ -23,9 +28,18 @@ const slice = createSlice({
   name: 'websocket',
   initialState,
   reducers: {
-    connected: (state) => {
-      state.connected = true;
-      delete state.error;
+    connected: {
+      reducer: (state, _action: PayloadAction<websocketConnectedPayload>) => {
+        state.connected = true;
+        delete state.error;
+      },
+
+      prepare: () => ({
+        payload: {
+          iAcceptThisIsAnUnsupportedApi: true,
+        },
+        meta: makeWebSocketMeta(),
+      }),
     },
 
     disconnected: (state) => {
@@ -48,6 +62,11 @@ export const {
   error: websocketError,
   featureFlagEnabled: websocketFeatureFlagEnabled,
 } = slice.actions;
+
+export const websocketConnectedSchema = createWebsocketResponseSchema(
+  websocketConnected,
+  websocketConnectedPayloadSchema,
+);
 
 export const websocketErrorSchema = createWebsocketResponseSchema(
   websocketError,
--- a/ui/frontend/websocketMiddleware.ts
+++ b/ui/frontend/websocketMiddleware.ts
@@ -4,12 +4,14 @@ import { z } from 'zod';
 import { wsExecuteResponseSchema } from './reducers/output/execute';
 import {
   websocketConnected,
+  websocketConnectedSchema,
   websocketDisconnected,
   websocketError,
   websocketErrorSchema,
 } from './reducers/websocket';
 
 const WSMessageResponse = z.discriminatedUnion('type', [
+  websocketConnectedSchema,
   websocketErrorSchema,
   wsExecuteResponseSchema,
 ]);
@@ -52,6 +54,7 @@ export const websocketMiddleware =
   (window: Window): Middleware =>
   (store) => {
     let socket: WebSocket | null = null;
+    let wasConnected = false;
     let reconnectAttempt = 0;
 
     let timeout: number | null = null;
@@ -75,15 +78,16 @@ export const websocketMiddleware =
         resetTimeout();
 
         socket.addEventListener('open', () => {
-          store.dispatch(websocketConnected());
-          reconnectAttempt = 0;
+          if (socket) {
+            socket.send(JSON.stringify(websocketConnected()));
+          }
         });
 
         socket.addEventListener('close', (event) => {
           store.dispatch(websocketDisconnected());
 
           // Reconnect if we've previously connected
-          if (!event.wasClean) {
+          if (wasConnected && !event.wasClean) {
             reconnect();
           }
         });
@@ -100,6 +104,12 @@ export const websocketMiddleware =
           try {
             const rawMessage = JSON.parse(event.data);
             const message = WSMessageResponse.parse(rawMessage);
+
+            if (websocketConnected.match(message)) {
+              wasConnected = true;
+              reconnectAttempt = 0;
+            }
+
             store.dispatch(message);
             resetTimeout();
           } catch (e) {
--- a/ui/frontend/websocketMiddleware.ts
+++ b/ui/frontend/websocketMiddleware.ts
@@ -52,7 +52,6 @@ export const websocketMiddleware =
   (window: Window): Middleware =>
   (store) => {
     let socket: WebSocket | null = null;
-    let wasConnected = false;
     let reconnectAttempt = 0;
 
     let timeout: number | null = null;
@@ -77,17 +76,14 @@ export const websocketMiddleware =
 
         socket.addEventListener('open', () => {
           store.dispatch(websocketConnected());
-
-          wasConnected = true;
+          reconnectAttempt = 0;
         });
 
         socket.addEventListener('close', (event) => {
           store.dispatch(websocketDisconnected());
 
           // Reconnect if we've previously connected
-          if (wasConnected && !event.wasClean) {
-            wasConnected = false;
-            reconnectAttempt = 0;
+          if (!event.wasClean) {
             reconnect();
           }
         });
@@ -114,15 +110,10 @@ export const websocketMiddleware =
     };
 
     const reconnect = () => {
-      if (socket && socket.readyState == socket.OPEN) {
-        return;
-      }
-
-      connect();
-
       const delay = backoffMs(reconnectAttempt);
       reconnectAttempt += 1;
-      setTimeout(reconnect, delay);
+
+      window.setTimeout(connect, delay);
     };
 
     connect();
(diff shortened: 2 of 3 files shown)


shepmaster:
I tested this locally by hackily slowing down the WebSocket connection code:

```diff
diff --git a/ui/frontend/websocketMiddleware.ts b/ui/frontend/websocketMiddleware.ts
index 8a540a4d..281b417a 100644
--- a/ui/frontend/websocketMiddleware.ts
+++ b/ui/frontend/websocketMiddleware.ts
@@ -12,6 +12,11 @@ import {
 
 const WSMessageResponse = z.discriminatedUnion('type', [WebSocketError, WSExecuteResponse]);
 
+function syncSleep(t: number) {
+  const start = Date.now();
+  while (Date.now() - start < t);
+}
+
 const reportWebSocketError = async (error: string) => {
   try {
     await fetch('/nowebsocket', {
@@ -68,6 +75,7 @@ export const websocketMiddleware =
     };
 
     const connect = () => {
+      syncSleep(200);
       socket = openWebSocket(window.location);
       if (socket) {
         resetTimeout();
```

The previous code could actually end up in a case where it seemingly would connect new WebSockets *forever* (or until resources ran out)

aguspdana:
> ```ts
> +      syncSleep(200);
> ```

I don't think it makes a difference here.

> The previous code could actually end up in a case where it seemingly would connect new WebSockets forever (or until resources ran out)

Do you mean the code before this PR? Yes, if the handshake takes more than 10 seconds. It stops creating a new WS connection until the last one is connected.

---

There's a WS gotcha that can't be debugged locally. When a connection is broken, the browser may not fire the close event immediately, it optimistically waits for the connection to get back to normal again. Usually the server sends ping/pong frame to ensure the connection is still alive. After `n` failed ping/pong, it may send a `close` frame. The client may never get the close message because the network is broken. In this case, we stuck with a WS connection that seems still connected but is actually not.

I'm not sure if this optimistic behavior can happen when the WS never reaches "open" state. If it does, with the current PR, the re-connection may never succeed because it attempts to reconnect only when the close event is fired.

---

```ts
if (!event.wasClean) {
  reconnect();
}
```

Continuing the above scenario. If the connection gets back to normal again, the client would receive the close message. Thus the close event would be fired with `event.wasClean === true`. In this case, the client would not reconnect, even though it should.

I think we should always reconnect on close unless `ws.close()` is called or when the server explicitly tells not to reconnect.

---

An alternative solution I have in mind is

```ts
...

// https://exponentialbackoffcalculator.com
const backoffMs = (n: number) => Math.min(5000 * Math.pow(2, n), 60000) * (1 + Math.random() / 2);

const idleTimeoutMs = 60 * 60 * 1000;

export const websocketMiddleware =
  (window: Window): Middleware =>
  (store) => {
    let socket: WebSocket | null = null;
    let reconnectAttempt = 0;

    let timeout: number | null = null;
    const resetTimeout = () => {
      if (timeout) {
        window.clearTimeout(timeout);
      }

      timeout = window.setTimeout(() => {
        if (!socket) {
          return;
        }

        socket.close();
      }, idleTimeoutMs);
    };

    const connect = () => {
      const alreadyConnected = socket && socket.readyState == socket.OPEN;
      
      if (alreadyConnected) {
        return;
      }
      
      socket = openWebSocket(window.location);
      
      if (socket) {
        resetTimeout();

        socket.addEventListener('open', () => {
          store.dispatch(websocketConnected());
          reconnectAttempt = 0;
        });

        socket.addEventListener('close', (event) => {
          store.dispatch(websocketDisconnected());

          // TODO: Always reconnect unless socket.close() is called or the server
          // explicitly tells to not reconnect.
          // Since the WS instance is not passed outside of the parent function
          // we reconnect uncondionally.
          connect();
        });

        socket.addEventListener('error', () => {
          // We cannot get detailed information about the failure
          // https://stackoverflow.com/a/31003057/155423
          const error = 'Generic WebSocket Error';
          store.dispatch(websocketError(error));
          reportWebSocketError(error);
        });

        socket.addEventListener('message', (event) => {
          try {
            const rawMessage = JSON.parse(event.data);
            const message = WSMessageResponse.parse(rawMessage);
            store.dispatch(message);
            resetTimeout();
          } catch (e) {
            console.log('Unable to parse WebSocket message', event.data, e);
          }
        });
      }
      
      const delay = backoffMs(reconnectAttempt);
      reconnectAttempt += 1;
      setTimeout(connect, delay);
    };

    connect();

    return (next) => (action) => {
      if (socket && socket.readyState == socket.OPEN && sendActionOnWebsocket(action)) {
        const message = JSON.stringify(action);
        socket.send(message);
        resetTimeout();
      }

      next(action);
    };
  };

const sendActionOnWebsocket = (action: any): boolean => action.type === ActionType.WSExecuteRequest;
```
Plus the client should initiate ping/pong, otherwise it won't detect hanging connection as I described above.

A better approach is to create a WS abstraction that handles the re-connection.

shepmaster:
I tried to follow your comment, but I'm not confident that I did so; my attempts at responses are below. Would you please also restate if there would be a concrete problem with the code if this PR was merged (and if so what to do to trigger the problem)?

Thanks!

---

> After 200 ms the previous connection will eventually connected

But in that case the `reconnect` function should exit early due to the`if (socket && socket.readyState == socket.OPEN)` guard. I totally can see how we could start a few reconnection attempts without waiting for the previous attempt to either succeed or fail. What I don't understand is what sequence of code causes the synchronous delay to spiral into an infinite number of connection attempts.

> There's a WS gotcha that can't be debugged locally.

This seems like a normal networking problem and there's not much to be done about it. Adding ping/pong doesn't solve the problem because the disconnection can happen in between the ping interval. All the ping/pong can do is decrease the time to notice the disconnection if nothing else is happening. You could increase the frequency at the expense of additional network traffic.

> Usually the server sends ping/pong frame to ensure the connection is still alive

We do not, so I don't think the hypothetical is currently relevant.

> or when the server explicitly tells not to reconnect.

Isn't that the same problem you previously described? The server notices the transient network issue, sends a "please do not reconnect" message, then the connection re-establishes, the client disconnects... right?

> A better approach is to create a WS abstraction that handles the re-connection.

This file _is_ the abstraction. I'm open to concrete suggestions on how to reorganize it though.

shepmaster:
One interesting difference is that in production, we have multiple proxies in the middle!

```
Browser -> CloudFlare -> nginx -> Axum
```

The proxies work by accepting the connection immediately, then sending the data upstream. In our case, nginx was the broken link, but connections to CloudFlare continued to work. That means that **the `open` event actually fired**, resetting our `reconnectAttempt` back to zero, but then we immediately closed and started the reconnection dance again.

To reproduce this locally, I proxied through [ProxyMan](https://proxyman.io) and disabled the `/websocket` endpoint. The problem occurred right away. Turning off the proxy resulted in a single failure.

shepmaster:
That proxy problem still occurs with the code proposed by this PR. It's *better* as now we don't get quite as many concurrent requests, but each request is pretty quickly after the previous.

My thoughts are:

1. Make the timeout / attempt counter a little bit more global. If we waited 10 seconds, then got a connection and closed, should we really only wait 1 ms to reconnect instead of 10 s again?
2. We should have some kind of handshake with the server to verify that we've really-really connected. I kind of wanted this for other unrelated reasons, but the response to that would be a good sign that the socket is up and running.

shepmaster:
Alright, this PR has grown a bit in size as I've brought in some refactoring I did in another branch; only the last two commits are directly relevant to the WebSocket reconnection problem. Since I needed to do a bit more communication, I didn't want to write something here that I'd have to rewrite in a day or two.

The penultimate commit addresses the original issue of performing overlapping reconnection attempts by only reconnecting after receiving the close, not just at an arbitrary interval. 

The last commit addresses the proxy issue by sending a message from the browser to the server and waiting for the response. Only once that is received will we perform reconnection attempts.

aguspdana:
> After 200 ms the previous connection will eventually connected

My bad. I thought `reconnectAttempt = 0;` was inside `open` listener.

> > There's a WS gotcha that can't be debugged locally.
>
> This seems like a normal networking problem and there's not much to be done about it. Adding ping/pong doesn't solve the problem because the disconnection can happen in between the ping interval. All the ping/pong can do is decrease the time to notice the disconnection if nothing else is happening. You could increase the frequency at the expense of additional network traffic.

You're right, it doesn't solve this problem. But eventually we need to implement it to avoid bad UX, e.g. the front-end stuck loading.

> or when the server explicitly tells not to reconnect.

This is only a hypothetical scenario, in case we want to limit the concurrent connection to prevent the server from running out of resources.

> This file is the abstraction. I'm open to concrete suggestions on how to reorganize it though.

I've created this gist you may want to check.
https://gist.github.com/aguspdana/987a7ee9f587087969fba4ea6a123f8d

The exposed methods are compatible with the standard WebSocket. Plus it has options to reconnect on close/error and queue messages while the socket is connecting. The reconnect delay is similar to this PR: delta between close event and reconnecting.

shepmaster:
> But eventually we need to implement it to avoid bad UX, e.g. the front-end stuck loading.

I'd expect that to be driven by the client, not the server. Additionally, it may be better served at a higher level. For example, an execute request will result in N stderr / stdout responses. If one of those doesn't return frequently enough, we'd want to reflect that in the UI.

> in case we want to limit the concurrent connection

Ah. I think that'd be covered in the current implementation — the server just will drop the handshake. Since it never finished completely, the client won't reconnect.

> Plus it has options to [...] queue messages while the socket is connecting

This might be something we want someday, but I don't think today. The playground UI likely will never rely solely on WebSockets, so we will always need to perform some check somewhere if the WebSocket is even available to be used. To know if it can be used, we will have needed to connect. The only time we'd need queuing would be during a reconnection scenario.

shepmaster:
Would you please also restate if there would be a concrete problem with the code if this PR was merged (and if so what to do to trigger the problem)?

aguspdana:
```ts
        ...
        socket.addEventListener('open', () => {
          if (socket) {
            socket.send(JSON.stringify(websocketConnected()));
          }
        });
        ...
        socket.addEventListener('message', (event) => {
          ...
            if (websocketConnected.match(message)) {
              wasConnected = true;
              reconnectAttempt = 0;
            }
          ...
        });
        ...
```

It's not necessary to exchange a message to ensure that the client and the
server are connected (`wasConnected = true`).  That's what the `open` event is for.
In this PR, if the connection is broken imidiately after it opens, before the
client gets the response back from the server, the client would never reconnect.

shepmaster:
Have you had a chance to read https://github.com/rust-lang/rust-playground/pull/912#issuecomment-1483414855 or the comment of the last commit?

aguspdana:
Sorry, I didn't give much thought for it before.

> I proxied through [ProxyMan](https://proxyman.io/) and disabled the /websocket endpoint. The problem occurred right away. Turning off the proxy resulted in a single failure.

So, you're trying to connect to a path that doesn't exist and the `open` event fired? What was the status code?

This is really weird.

I visited the playground and ran this in the console:

```js
(function () {
    let socket = new WebSocket("wss://play.rust-lang.org/websocket-not-exist");
    socket.onopen = function () { console.log("open") };
    socket.onclose = function () { console.log("close") };
})()
```

Only the `close` event was fired. And the browser only showed provisional header.

```
Provisional headers are shown
```

Could that be that there's a bug in ProxyMan?

shepmaster:
I didn't see anything with a local Nginx reverse proxy, so that's probably a dead end of investigation. I'm going to go ahead and merge this because the previous behavior is clearly wrong and this appears to be better, even if it may not fix every possible case.

Issue #626, opened 2020-08-11
Add a notice on top of /etc/passwd that the playground is sandboxed


pietroalbini:
The Rust Security Response WG occasionally receives security reports about remote code execution on the playground. Since the playground uses a sandbox to mitigate that, most of the reports are actually not a security vulnerability.

To reduce the workload of the Security Response WG, and to save potential security researchers time, it would be nice to add a comment on top of `/etc/passwd` and `/etc/shadow` like this:

```
# Hello, and thanks for looking into the Rust Playground's security!
#
# This build is running on an unprivileged, sandboxed Docker container with no
# network access, so while you can technically run arbitrary code on the
# Playground you shouldn't be able to do any damage with it.
#
# Nothing is perfect though: if you find a way to escape the sandbox, please
# disclose it following our security policy! You can find the policy at:
#
#    https://www.rust-lang.org/policies/security
#
```

Those two files are often used by security researcher to demonstrate RCEs, so adding the notice on top of those two files should be enough!

0x4ndy:
Would it make sense to include this change in `rust-playground/compiler/base/Dockerfile`?

shepmaster:
Great question! I think `postinstall.sh` would be a reasonable place as you could then make a little function that takes in the filename as an argument to avoid duplicating the content.

0x4ndy:
That's my thinking precisely. I've just made a fork, so I could do it and make pull request.

0x4ndy:
Unfortunately adding the change in the `postinstall.sh` wasn't feasible due to the fact that the script is running as a `playground` user. Therefore, I've created a dedicated script and added it to the Dockerfile. 

The script is fairly generic so that it's possible now to add any text file to any other text file, either on top or on the bottom, based on the arguments passed during the execution (just in case a similar need arie in the future for another use case). 

Here's the commit: 
https://github.com/rust-lang/rust-playground/commit/7824753c7ede8c752042257e1e0878b48a5a924d

Please let me know if that's good enough, I will then make a Pull Request.

PR #925, opened 2023-04-24
Add a security notice to commonly inspected files

0x4ndy:
I've added a fairly generic script that allows to include any text file to any other text file, either on top or on the bottom, based on the arguments passed during the execution (just in case a similar need arie in the future for another use case). 

I've also updated the Dockerfile to execute the script at the end (before the entrypoint). Since the intention was to modify `/etc/passwd` and `/etc/shadow` files, I had to change the `USER` to `root` temporarily. After the files are updated, I changed the user back to `playground`.

The security notice I copied as is from #626. 

The updates are done from `/tmp` folder and at the end the script is removed. However, I decided to leave the security notice itself in the `playground` folder: `/playground/security_notice.txt` so that as soon as someone enumerates the content of the current directory, they will see the file.

### DIFF
--- a/compiler/base/Dockerfile
+++ b/compiler/base/Dockerfile
@@ -20,6 +20,13 @@ RUN apt-get update && apt-get install -y \
 
 RUN useradd -m playground -d /playground
 RUN usermod -p '!!' root # Disable all passwords for root
+
+# Attach the security note
+ADD --chown=playground attach_notice.sh security_notice.txt /playground/
+RUN /playground/attach_notice.sh /playground/security_notice.txt /etc/passwd && \
+    /playground/attach_notice.sh /playground/security_notice.txt /etc/shadow && \
+    rm -f /playground/attach_notice.sh
+
 USER playground
 ENV USER=playground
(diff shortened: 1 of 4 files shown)


shepmaster:
I've [force-pushed changes](https://github.com/rust-lang/rust-playground/compare/7824753c7ede8c752042257e1e0878b48a5a924d..2cf67a87ba8c0305dca5bfda08c5a5f659ac8e25). A brief list:

- Reworded commit message. Stating that a commit fixes an issue is useful but not sufficient. Ideally the `git log` can describe the changes in a standalone fashion, linking out for further background information.
- Moved the Dockerfile steps earlier to avoid switching users
- Reduced the number of Dockerfile steps to reduce the number of intermediate layers
- Preserved the permissions of the original file when we add the notice at the top.
- Used `mktemp` to generate a unique filename instead of hardcoding it.
- Quoted variables like `"${foo}"`. This is mostly a pedantic thing to help handle cases like values with spaces.
- Added an automated test

I'll soon rebase this on some upcoming maintenance changes and then let CI take a crack at it.

PR #962, opened 2023-09-05
Clarify suggestion to use extern crate in help

schuelermine:
`extern crate` is no longer necessary in most cases since Rust 2018

This PR adds a clarification to the help page


shepmaster (approved):
This is great and has bothered me whenever I notice it (and then promptly forget to fix it).

I'm going to go ahead and merge this, but if you want to tweak it even further we could flip the wording to put the common case (2018+) first. Something like "add `use foo` (and if using Rust 2015, also add an `extern crate foo`"

schuelermine:
Just noticed this is missing a space, see #967

PR #967, opened 2023-09-07
Fix mistake in #962 that introduced a missing space

schuelermine:
In #962 I accidentally introduced a missing space, this hopefully fixes that


shepmaster (approved):
Ugh, JSX is super frustrating with that. Thanks!

PR #972, opened 2023-09-14
Prevent output pane text from being copied as HTML

### LINKED ISSUE #840: Make stderr output more copy-friendly
When copying the stderr output from the compiler, it seems the clipboard now contains markdown for the link around error codes in the main message (or maybe it's HTML and pasting in GH turns it into markdown). This is causing unnecessarily verbose output in reports. I believe that in most cases people desire just the textual representation of errors, the link structure being either redundant or unnecessary. I would love it if we could capture the on copy event and only push the bare textual output to the clipboard.

notriddle:
Fixes #840

### DIFF
--- a/ui/frontend/Output/SimplePane.tsx
+++ b/ui/frontend/Output/SimplePane.tsx
@@ -10,8 +10,19 @@ interface HighlightErrorsProps {
   label: string;
 }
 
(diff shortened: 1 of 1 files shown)


shepmaster:
Thanks!

PR #978, opened 2023-09-19
add strum crate to base

alekspickle:
There was so many times I wanted to use strum to share demos with enums.
It's not heavy and I am sure more and more people will use it


shepmaster:
https://github.com/rust-lang/rust-playground/blob/main/CRATE_POLICY.md

PR #992, opened 2023-11-03
Update outdated links

kyu08:
Thanks for always maintaining rust-playground!
I found some outdated links and have corrected them.


shepmaster (approved):

shepmaster:
Thanks!

PR #994, opened 2023-11-06
Support Rust edition 2024

### LINKED ISSUE #993: No edition 2024 in the combobox.
How do I play with [gen blocks](https://github.com/rust-lang/rust/pull/116447), which seem to require setting edition to unstable so far 2024? Even manually editing `edition=2021` URL parameter does not make it force the new edition.

Maybe 2024 should be added in GUI for Nightly and custom editions (e.g. 2027) be specifiable in the URI?

shepmaster:
Closes #993

### DIFF
--- a/compiler/base/modify-cargo-toml/src/lib.rs
+++ b/compiler/base/modify-cargo-toml/src/lib.rs
@@ -26,6 +26,22 @@ fn ensure_string_in_vec(values: &mut Vec<String>, val: &str) {
     }
 }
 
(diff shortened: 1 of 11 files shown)


Skgland:
I noticed that this automatically switches to nightly when edition 2024 is selected, but if edition 2024 is selected and one switches from nightly to beta or stable edition 2024 stays selected. Should it change to the latest stable i.e. edition 2021 instead so that a valid channel edition combination is selected?

shepmaster:
I wouldn’t be sad if someone made it do that, but I’m not overly worried about it.

PR #1023, opened 2023-12-25
Add dark mode

jprochazk:
Fixes https://github.com/rust-lang/rust-playground/issues/544

This uses the media query approach described in the issue. The dark theme I came up with here was created using the [dark reader](https://darkreader.org/) extension and some manual tweaking. I'm not a designer, so happy to change the colors further until it looks acceptable.

# `prefers-color-scheme: dark`
image

# `prefers-color-scheme: light` (unchanged)
image

### DIFF
--- a/ui/frontend/.stylelintrc.json
+++ b/ui/frontend/.stylelintrc.json
@@ -12,6 +12,9 @@
       "ignoreProperties": [
         "composes"
       ]
+    }],
+    "at-rule-no-unknown": [true, {
+      "ignoreAtRules": ["define-mixin", "mixin"]
     }]
   }
 }
--- a/ui/frontend/BuildMenu.module.css
+++ b/ui/frontend/BuildMenu.module.css
@@ -1,4 +1,4 @@
 .code {
-  background: #eee;
+  background: var(--code-background-color);
   padding: 0 0.25em;
 }
--- a/ui/frontend/ButtonSet.module.css
+++ b/ui/frontend/ButtonSet.module.css
@@ -76,11 +76,7 @@ $radius: 4px;
 
 .secondary {
   composes: -border -button;
-  background: linear-gradient(
-    to bottom,
-    var(--button-secondary-bg-color-top),
-    var(--button-secondary-bg-color-bottom)
-  );
+  background: var(--button-secondary-bg-color);
   border-color: var(--button-secondary-border-color);
   color: var(--button-secondary-color);
 
@@ -121,4 +117,9 @@ $radius: 4px;
   padding: 0;
   aspect-ratio: 1/1;
   justify-items: center;
+}
+
+a.iconLink svg,
+a.iconLink:visited svg {
+  fill: var(--font-color) !important;
 }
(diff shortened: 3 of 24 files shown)


shepmaster:
Thanks, this is a great start! 

Like you, I'm not a designer and I think it's fine to iterate on specific colors after getting all the CSS variables in place. There are two main points I have with the current state:

1. There are more colors that seem like they should be updated. Key ones:
  - Notifications 
      image
  - Menus 
        image
  - Links (these are basically unreadable)
      image
  - Help page (and code samples)
      image

2. Making the color scheme choosable by the user. I think that basically we should have a user setting that is { `light`, `dark`, `system` } that defaults to `system`.

I think point 1 should be addressed in this PR. Point 2 can be done later, I'd just like to make sure that we can see a path forward to that world.

jprochazk:
> 1. There are more colors that seem like they should be updated

I went through all the ones you listed, and found a few more along the way, such as the `<select>` and `<option>` element background colors.

> 2. Making the color scheme choosable by the user

I have a PoC for that on this branch: https://github.com/jprochazk/rust-playground/tree/dark-mode-selector

It adds a `website.theme` configuration option, and a selector for it in the `Config` menu. The configuration option is persisted in local storage. The theme is now controlled by both JS and the media query:

- If the configuration is set to `system`, then the theme depends on the `prefers-color-scheme` media query. If the browser does not support it, it defaults to light mode.
- Otherwise the theme is set to `light` or `dark` depending on the chosen configuration value.

To avoid duplicating all the theme variables, I added the `postcss-mixins` plugin and defined the variables in mixins.

shepmaster:
Howdy, long time no talk! (sorry about that)

I'd love it if you could take a [look at my branch `dark-mode-mods`](https://github.com/rust-lang/rust-playground/tree/dark-mode-mods). I've taken your PR and done a bit more tweaking, squashing some commits, and pulled in your other branch to add the config option.

On top of all of that, I've also hidden all your hard work behind a feature flag 😬. I did this because I want to stop my faffing about and get your work out there and get some initial feedback, then turn it on for everyone. If you run the code locally, you can turn on the feature flag by adding `&features=true` to the URL. This *won't* turn on the automatic detection from `prefers`, but that's a simple enough part that I'm OK testing that live.

If my changes look good, I'll push it to your fork and merge things.

jprochazk:
> I'd love it if you could take a [look at my branch `dark-mode-mods`](https://github.com/rust-lang/rust-playground/tree/dark-mode-mods?rgh-link-date=2024-09-15T01%3A13%3A01Z).
> ...
> On top of all of that, I've also hidden all your hard work behind a feature flag 😬. 

Thanks for taking a look! I'm glad we can move this forward, and it makes a lot of sense to put this behind a feature flag.

One thing I ran into is that when I set `features=true`, the theme dropdown is set to `Dark` on the first load after a `localStorage.clear()`, but the website isn't actually in dark mode:

image

I have to choose `Light` and then `Dark` again to get dark mode.

Other than that, the changes look good to me!

shepmaster:
Thanks!

PR #1056, opened 2024-04-25
fix: use kebab-case as underscores are not allowed in edition 2024

weihanglo:
The change is compatible. Just prepare for edition 2024 in advance.

See also:

* <https://github.com/rust-lang/cargo/issues/13629>
* <https://github.com/rust-lang/cargo/pull/13783>
* <https://github.com/rust-lang/cargo/pull/13798>
* <https://github.com/rust-lang/cargo/pull/13804>


shepmaster:
Thanks!

PR #1072, opened 2024-07-04
Change `asm-comments` to `verbose-asm`

tgross35:
This option was recently renamed, so the playground's assembly output is broken on the nightly channel. See <https://github.com/rust-lang/rust/pull/126803#event-13384747335>.


shepmaster (approved):
Thanks!

PR #1074, opened 2024-07-22
Allow user to specify aliasing model for Miri

GoldsteinE:
Very useful for comparing stacked borrows and tree borrows.

An example for testing:
```rust
fn main() {
    let val = [1u8, 2];
    let ptr = &val[0] as *const u8;
    let _val = unsafe { *ptr.add(1) };
}
```
is valid under Tree Borrows, but not under Stacked Borrows.

UI looks like this:

image

### DIFF
--- a/compiler/base/cargo-miri-playground
+++ b/compiler/base/cargo-miri-playground
@@ -3,5 +3,5 @@
 set -eu
 
 export MIRI_SYSROOT=~/.cache/miri
-export MIRIFLAGS="-Zmiri-disable-isolation"
+export MIRIFLAGS="${MIRIFLAGS:-} -Zmiri-disable-isolation"
 exec cargo miri run
--- a/compiler/base/orchestrator/src/coordinator.rs
+++ b/compiler/base/orchestrator/src/coordinator.rs
@@ -239,6 +239,12 @@ pub enum Channel {
     Nightly,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq)]
+pub enum AliasingModel {
+    Stacked,
+    Tree,
+}
+
 impl Channel {
     #[cfg(test)]
     pub(crate) const ALL: [Self; 3] = [Self::Stable, Self::Beta, Self::Nightly];
@@ -654,6 +660,7 @@ pub struct MiriRequest {
     pub channel: Channel,
     pub crate_type: CrateType,
     pub edition: Edition,
+    pub aliasing_model: AliasingModel,
     pub code: String,
 }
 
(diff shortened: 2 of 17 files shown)


GoldsteinE:
The one problem I have with my implementation is that this knob is not saved when sharing via link. I could easily fix this, but it will add an extra `&aliasing_model=stacked` at the end of every single URL. Alternatively, I could make “Share” omit unchanged parameters, but I suppose there’s a reason why it isn’t this way already.

GoldsteinE:
@shepmaster Hi! Looking at the commit history, I think I should ask you to review these changes.

shepmaster:
> this knob is not saved when sharing via link

I think that's fine for now. Similarly, the "backtrace" configuration is not added as a query parameter.

> I suppose there’s a reason why [“Share” does not omit unchanged parameters]

That's basically to avoid dealing with changing defaults. For example, if you click on a playground link from the year 2015, it will not have the `edition` query parameter since editions did not exist yet. That means that the value for a missing edition needs to forever be `2015`. Moving forward, we always add the edition parameter even when the edition matches the current default, as the default changes over time.

In this case, Miri defaults to stacked borrows today, but conceptually could switch to tree borrows tomorrow. A URL generated today with no value would become ambigous if loaded tomorrow.

---

All that said, I've been mulling over a richer way of saving data about the current playground instance. If enough interest arises from people missing this setting sticking around, I might look into that.

shepmaster:
Thanks for your patience! The code looks good, but I did pull it down and make some small tweaks, as well as rebasing it to address some conflicts. I then [force-pushed those changes](https://github.com/rust-lang/rust-playground/compare/be78b11b1da6256617badd9f6a1dc8bc1f6c6260..0e65195d5aa78eef18eb4ebc7685eec59d0cc93a). Due to that, your two commits are no longer verified. If that's OK with you, I can go ahead and merge this. Otherwise, feel free to review the commits then rebase them and force-push yourself with the new approvals.

GoldsteinE:
> All that said, I've been mulling over a richer way of saving data about the current playground instance. If enough interest arises from people missing this setting sticking around, I might look into that.

The thing I find myself missing the most is the state of the big “Run” button: I often want to share some code in ASM or LLVM IR mode and right now the best way to do it is to just specify it in prose. (Aside: I think “Expand macros”, “Miri” and “Clippy” could also be under the “Run” button instead of in “Tools”). I don’t think not saving Miri mode is that bad, since it’s mostly useful for _comparing_ Miri modes, so you’ll need to switch it anyway.

> Due to that, your two commits are no longer verified. If that's OK with you, I can go ahead and merge this. Otherwise, feel free to review the commits then rebase them and force-push yourself with the new approvals.

Give me a minute, I’ll review and re-sign these commits.

GoldsteinE:
I signed the commits (adding you as a co-author for the second one).

RalfJung:
We usually spell the tool's name "Miri", would be good to be consistent here as well. :)

RalfJung:
Ah, seems like the section titles are generally all-caps... I guess this will lead to more people spelling it MIRI but oh well.

shepmaster:
> Ah, seems like the section titles are generally all-caps

If it makes you feel any better, I do spell it the right way in the pre-stylized code 😉 :

```jsx
<MenuGroup title="Miri">
```

I feel your pain though, albeit in the opposite direction. I generally prefer [SNAFU](https://docs.rs/snafu/latest/snafu/index.html) to be in all caps as it's an acronym, but most people pick something else.

shepmaster:
[Tests passed](https://github.com/rust-lang/rust-playground/actions/runs/13657556260?pr=1074)

shepmaster:
Thank you for your patience.

RalfJung:
Thank you for pushing this over the finish line :)

PR #1083, opened 2024-09-06
Display Axum server IP and Port info with `RUST_LOG=info`

bsodmike:
When running the server, I noticed IP/port info of the Axum server were not shown.

Running `RUST_LOG=info` will show this now.

### DIFF
--- a/ui/src/server_axum.rs
+++ b/ui/src/server_axum.rs
@@ -156,7 +156,10 @@ pub(crate) async fn serve(config: Config) {
         MakeRequestUuid::default(),
     ));
 
-    let listener = tokio::net::TcpListener::bind(config.server_socket_addr())
+    let server_socket_addr = config.server_socket_addr();
+    tracing::info!("Serving playground backend at http://{server_socket_addr}");
+
+    let listener = tokio::net::TcpListener::bind(server_socket_addr)
         .await
         .unwrap();
 


ui/src/server_axum.rs
```diff
@@ -160,6 +160,19 @@ pub(crate) async fn serve(config: Config) {
...
+        .local_addr()
+        .expect("Unable to fetch local `SocketAddr`");
+    tracing::info!(
+        "Serving playground backend at {}:{} [{}]",
+        &socket_addr.ip(),
+        &socket_addr.port(),
+        (match &socket_addr.is_ipv4() {
+            true => "IPv4",
+            _ => "IPv6",
+        })
```
shepmaster, line 173: Can you explain a little bit more about why you want to know this information? IPv4 and IPv6 addresses are very visually different. IPv4 looks like `127.0.0.1:5000` and IPv6 looks like `[::]:6000`. In what case would a person need to know if it's IPv4 vs IPv6 but not already know the difference in formatting?

bsodmike: Hmm good point.  I've seen some (other) services sometimes starting up on an ipv6 interface and this has the potential of confusing someone who isn't aware, but as you pointed out, that would be unlikely.

I'll make the update, thanks.

shepmaster:
Alright, after dealing with some upstream annoyances, I rebased this on some updates and made some tweaks to remove the error condition, simplify the printing, and include `http://` which can make the information clickable in many terminals. Still look good to you?

bsodmike:
> Alright, after dealing with some upstream annoyances, I rebased this on some updates and made some tweaks to remove the error condition, simplify the printing, and include `http://` which can make the information clickable in many terminals. Still look good to you?

Perfect. Shift + clicking in Zellij works great for me too.  Thanks!

shepmaster (approved):

PR #1130, opened 2025-01-02
Increasing zerocopy (and zerocopy-derive) version from 0.7.35 to 0.8.14

peter-lyons-kehl:
Existing import of https://docs.rs/zerocopy/0.7.35/zerocopy/ is superseded by https://docs.rs/zerocopy/0.8.14/zerocopy (the current). There are (probably/hopefully only) minor incompatibilities.

Current (0.8.14-based) defines new functionality (`KnownLayout`, `Immutable` traits and their derives). There is no value in importing the old version and keeping any old gists/examples working, since that will only confuse any newcomers. Actually, if people notice that their existing gists/examples fail, hopefully they will have more heads up time to migrate.

Feel free to ask Zerocopy maintainers @joshlf and @jswrenn.

Dear Zerocopy maintainers, if you have a "Release"/major API change checklist (even though anything before 1.0.0 is considered unstable, of course), suggest updating/notifying rust-playground, too.

Thank you in advance to all of you.

### DIFF
--- a/compiler/base/Cargo.toml
+++ b/compiler/base/Cargo.toml
@@ -1817,12 +1817,12 @@ version = "=0.7.5"
 
 [dependencies.zerocopy]
 package = "zerocopy"
-version = "=0.7.35"
+version = "=0.8.14"
 features = ["__internal_use_only_features_that_work_on_stable", "alloc", "byteorder", "derive", "simd", "zerocopy-derive"]
 
 [dependencies.zerocopy_derive]
 package = "zerocopy-derive"
-version = "=0.7.35"
+version = "=0.8.14"
 
 [dependencies.zerofrom]
 package = "zerofrom"


shepmaster:
Please read [our crate policy][policy]. Any issue opened for a new or updated crate version will be closed with a reference to the same policy.

[policy]: https://github.com/rust-lang/rust-playground/blob/master/CRATE_POLICY.md

shepmaster:
All crate versions will get updated automatically at some point so long as the crate is still used.

peter-lyons-kehl:
Thank you Jake and Carol for your contribution to Rust and our community. Happy New Year.

I have read CREATE_POLICY.md and README.md. They don't say, nor imply, anything about if/when/how crate versions get updated (except for "temporarily excluded" in "Exclusion policy" section).

shepmaster:
Right. The first post is the boilerplate used when an issue / PR is opened with a new / updated crate version. The second post is further detail explaining when updates happen ("whenever I feel like it, but usually at least once per Rust version" is the closest truth at the moment).

peter-lyons-kehl:
The issue template boiler plate tells people off, as if they didn't read `CRATE_POLICY.md`. Many don't read it - but some do. Either way: What's the value in referring people (regarding updates) to `CRATE_POLICY.md`, if it doesn't mention updates?

However, on my 3rd reading of `CRATE_POLICY.md` > first section ("Playground crate inclusion policy") contains: "The **latest** stable version of these crates are available" (bold formatting is mine).

`zerocopy` is below `1.0`, so not stable, but still: "The latest... are available" vs. " "at some point" and "whenever...". Can you see the space for confusion?

I'm not questioning how often you update - to the contrary, I'm grateful. But, people are unhappy that they don't know what to expect, and (unhappy) about this policy not being clear and written in one place.

shepmaster:
> The latest stable version *of these crates* are available

"these crates" refers to the top 100 downloads and those in the cookbook. zerocopy is not in this list, so it is not guaranteed to have the latest stable version. It is only a dependency of another crate (looks like ahash) and thus only has what version is needed for that code to compile. You could encourage ahash to upgrade their dependency. If they do so, release a new version, and then we update, then zerocopy will be updated implicitly.

PR #1133, opened 2025-02-03
Update derive_more and include all features

JelteF:
The `derive_more` version in the playground was outdated and didn't include many of the features that the library provides. This fixes that. 

(I'm the original author of `derive_more` and I was very happy to find out it is available on the playground)

### DIFF
--- a/compiler/base/Cargo.toml
+++ b/compiler/base/Cargo.toml
@@ -356,9 +356,8 @@ version = "=2.2.0"
 
 [dependencies.derive_more]
 package = "derive_more"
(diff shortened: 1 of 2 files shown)


shepmaster:
TL;DR: If you want to control which features are available from a crate you own, see https://github.com/rust-lang/rust-playground/issues/192. Otherwise, the version of crates like `derive_more` (which I don't think is in the top 100 downloads) are controlled by whatever crate is pulling them in. You'll want to adjust those crates to pull in a new version.

PR #1138, opened 2025-02-14
Cancel tokens and abort tasks on drop

shepmaster:
When a client disconnects, Axum drops our future. Any tasks we've spawned need to be properly notified to also exit. Enhance tasks to be aborted and cancellation tokens to be triggered when dropped.

I hope that this helps clear up some of the cases where the playground thinks that a container is still running after it has exited.


shepmaster:
> I hope that this helps clear up some of the cases where the playground thinks that a container is still running after it has exited.

Narrator: It did not.

Issue #1172, opened 2025-08-24
Unable to run locally using docker


Souvlaki42:
I propably don't understand how to build this project locally in my Ubuntu VPS.

I ran the `fetch.sh` script and it fetches the containers just fine.

I ran the `build.sh` script which takes longer as expected and finishes just fine after modifying the `Dockerfile` to resolve some permission errors.

I build the frontend using nodejs and pnpm just fine.

In no point, I see any docker containers start up.

When trying to run the backend, I see errors because it tries to kill some docker containers that doesn't exist.

Am I doing something wrong or the instructions lack a section about actually running the containers needed?

Sorry in advance, if I'm dumb.

shepmaster:
> I ran the `fetch.sh` script 
> 
> I ran the `build.sh` script

You only need to do one or the other of these.

> In no point, I see any docker containers start up.

Did you access the playground through a browser and submit a request?

> kill some docker containers that doesn't exist.

Yes, that's a known annoyance. Normally I'd hide those warnings, but I'm in the middle of a (very long) debugging session so they are emitted at a higher level. Normally it doesn't matter. This arises because we try to kill the container even when it exits normally because there's generally no harm in making sure.

> the instructions lack a section about actually running the containers needed

You don't run any containers by hand — the playground does that for you.

PR #1174, opened 2025-09-03
Improved instructions & docker permissions fix

Souvlaki42:
This PR is a direct consequence of these two #1172 and #1173. 
It changes the following:
- Fixes some weird permission issues in the `Dockerfile` that runs with `compiler/build.sh`.
- Adds bash syntax highlighting in the shell code blocks in `README.md`.
- Adds [markdown highlights](https://github.com/orgs/community/discussions/16925) in `README.md` to improve readability.
- Adds a troubleshoot in `ui/README.md` about having to set the server's address to `0.0.0.0` to expose it to the world.
- Adds a section in `README.md` that highlights:
> [!IMPORTANT]
> When running the server, you may see some docker warnings about missing containers. You can safely ignore them. You shouldn't see any docker containers running most of the time. This is normal. You don't need to spawn or kill any containers yourself. The backend manages that itself.


Souvlaki42:
> I don't mind taking the documentation/style improvements, but I see no reason to change the Dockerfile. Notably, building the containers works fine on both my machine and in CI. At the moment, it seems more likely to me that you have something unique to your environment that needs to be figured out.

Oh, I'm more than happy to find out what misconfiguration I have to cause permission issues in Docker that get fixed by my first commit and don't happen in any other Dockerfile I've used. I don't say that to be rude, I genuinely feel that.

Do you have any idea?

Anyways, I care way more about the documentation/style improvements being accepted because I feel that they may help potential newcomer contributors and people like me who just want to play with it locally. Thank you for your response.

Souvlaki42:
> > Do you have any idea?
> 
> Nope, because you never stated what the errors were 😉. It may also be useful to share what version of Docker you are using.

Yeah sorry I forgot to mention that.

The first tests I did running rust-playground was in a Digital Ocean VPS running Ubuntu 25.10 with Docker template, while I was away and didn't have access to a real Linux machine.

It seems like it was a misconfiguration on their part or mine, but since these patches fixed those, I just included them just in case, in was a general issue.

A local run I'm doing right now, seems to go without issues even after I reverted those changes.

The only related logs I could find from that VPS said some like that:
```bash
Step 32/51 : RUN cargo chef prepare
---> Running in a1505111e42d
Error: Failed to save recipe to 'recipe.json'

Caused by:
0: failed to create file recipe.json
1: Permission denied (os error 13)
The command '/bin/sh -c cargo chef prepare' returned a non-zero code: 1

Error: Failed to cook recipe.

Caused by:
0: failed to create file /orchestrator/Cargo.lock
1: Permission denied (os error 13)

cargo run --release Error: The IO queue task failed

Caused by this error:
1: Failed to receive worker message
```

Anyways, I reverted the docker file changes.

PR #1241, opened 2026-07-14
Add some text when JS is disabled

GuillaumeGomez:
Currently, when JS disabled when we arrive on `play.rust-lang.org`, we just have an empty page. With this PR, when JS is disabled, it looks like this:

image

I made it very simple to convey the needed information, but I can add more content if you think it's needed.

### DIFF
--- a/ui/frontend/index.ejs
+++ b/ui/frontend/index.ejs
@@ -10,6 +10,9 @@
   </head>
   <main>
     <body>
+      <noscript>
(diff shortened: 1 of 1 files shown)


shepmaster:
Ugh, looks like GHA changed some default rules. I'll need to go figure out if there's a smarter way to handle the security here.

shepmaster:
Thanks!

GuillaumeGomez:
Didn't expect it would be so complicated. :laughing:

PR #1250, opened 2026-07-27
Ignore main functions inside block comments

### LINKED ISSUE #1167: `fn main` in multiline comment causes errors
Putting a `main` function in a multiline comment (this does not happen when using `//` commenting syntax on each line of `main`) causes the playground to mistakenly show a `Run` button rather than a `Build` button. This causes errors when the `Run` button is pressed.

```rust
/*
fn main() {}
*/
```
produces:
```
   Compiling playground v0.0.1 (/playground)
error[E0601]: `main` function not found in crate `playground`
 --> src/main.rs:3:3
  |
3 | */
  |   ^ consider adding a `main` function to `src/main.rs`

For more information about this error, try `rustc --explain E0601`.
error: could not compile `playground` (bin "playground") due to 1 previous error
```
This also causes a much more difficult to debug error message when you're trying to check that some functions work under `#![no_std]` and you comment out the `main` function:

```rust
#![no_std]

/*
fn main()
*/
```
produces:
```
   Compiling playground v0.0.1 (/playground)
error: `#[panic_handler]` function required, but not found

error: unwinding panics are not supported without std
  |
  = help: using nightly cargo, use -Zbuild-std with panic="abort" to avoid unwinding
  = note: since the core library is usually precompiled with panic="unwind", rebuilding your crate with panic="abort" may not be enough to fix the problem

error[E0601]: `main` function not found in crate `playground`
 --> src/main.rs:1:11
  |
1 | #![no_std]
  |           ^ consider adding a `main` function to `src/main.rs`

For more information about this error, try `rustc --explain E0601`.
error: could not compile `playground` (bin "playground") due to 3 previous errors
```

Guflly:
Block-commented main functions no longer select Run automatically.

This removes multiline and unclosed block comments before applying the existing main-function check. It leaves comment markers inside line comments, normal strings, and raw strings alone.

Fixes #1167

### DIFF
--- a/ui/frontend/selectors/index.spec.ts
+++ b/ui/frontend/selectors/index.spec.ts
@@ -59,10 +59,6 @@ describe('checking for a main function', () => {
     expect(hasMainFunction('/* fn hidden() {} */\nfn main() {}')).toBe(true);
   });
 
-  test('a main inside nested block comments does not count', () => {
-    expect(hasMainFunction('/* outer\n/* inner */\nfn main() {}\n*/')).toBe(false);
-  });
-
   test('block comment markers in strings do not hide a main', () => {
     expect(hasMainFunction('const marker: &str = "/*";\nfn main() {}')).toBe(true);
   });
--- a/ui/frontend/selectors/index.spec.ts
+++ b/ui/frontend/selectors/index.spec.ts
@@ -59,6 +59,10 @@ describe('checking for a main function', () => {
     expect(hasMainFunction('/* fn hidden() {} */\nfn main() {}')).toBe(true);
   });
 
+  test('a main inside a block comment after code does not count', () => {
+    expect(hasMainFunction('const X: () = (); /*\nfn main() {}\n*/')).toBe(false);
+  });
+
   test('block comment markers in strings do not hide a main', () => {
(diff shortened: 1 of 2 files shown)


shepmaster:
Thanks for the PR! I'm curious: what level of computer assistance did you have in creating this code? It's not *super* common for people to write out a parser from scratch.

I'm really quite torn on whether or not this is a good path to continue on, however. This code splits up the string character-by-character, does the logic, then creates the string again; that all feels overly wasteful. A quick benchmark using a 47K Rust file I have lying around shows that this code now takes ~2.8ms from the previous ~1.9ms. There's also a giant blob of new code to maintain. On top of that, it's all ultimately supporting regex hacks.

I whipped together a quick prototype of using tree-sitter compiled to Wasm, and that's even slower (~5.8ms), but at least that should be a very robust solution and could be used for multiple purposes.

Guflly:
I used an AI coding assistant for the first draft, then reviewed and tested it locally. You were right that the scanner was too much for this. I replaced it with a small regex-based pass that preserves line comments, normal strings, and raw strings. On a synthetic 47 KB input it took about 0.15 ms. The selector tests, lint, formatting, and type checking pass.

shepmaster:
See #1263 for my alternate solution.

shepmaster:
Closed via #1263

PR #1265, opened 2026-08-06
Move the editor's ruler to where rustfmt will wrap

### LINKED ISSUE #1264: Misleading mismatch between column guide (80 columns) and rustfmt (…95 columns? I think?)
play.rust-lang.org has a column guide (grey line) at 80 columns, but using Tools > rustfmt does not use 80 columns as the max width, which led to me misunderstanding what was going wrong in https://github.com/rust-lang/rustfmt/issues/7010. It would be nice™ if these two settings were in sync, so that the playground could better be used for testing rustfmt.

shepmaster:
This also turns on the ruler for Monaco in the first place.

We are picking the value from the current version of rustfmt, but I don't think it changes too often.

Closes #1264

### DIFF
--- a/ui/frontend/editor/AceEditorCore.tsx
+++ b/ui/frontend/editor/AceEditorCore.tsx
@@ -8,6 +8,7 @@ import React, { useCallback, useEffect, useMemo, useRef, useState } from 'react'
 
 import { CommonEditorProps, Crate, FileId, PairCharacters, Position } from '../types';
 import { useLatest } from './useLatest';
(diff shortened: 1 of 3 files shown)


jrose-signal:
You could also explicitly set the value in your rustfmt invocation to make sure it doesn't drift in the future!

shepmaster:
A fair point, but then I get in the business of defining a style guide, which I don't really want to do.

It's even seems possible someone would open an issue asking why the playground's UI and rustfmt's defaults don't match 😉

